Issue #401, opened 2014-10-23
Why don`t work params parity: 'even' and parity: 'odd'  FT232RL converter ?


Specialistvlad:
Don`t work params parity: 'even' and parity: 'odd'  FT232RL converter ?

uname -a 3.13.0-30-generic #55-Ubuntu SMP Fri Jul 4 21:40:53 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
node -v v0.10.32

Device: 
[48941.030399] usb 3-1: new full-speed USB device number 17 using xhci_hcd
[48941.053359] usb 3-1: New USB device found, idVendor=0403, idProduct=6001
[48941.053369] usb 3-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[48941.053374] usb 3-1: Product: USB Serial Converter
[48941.053379] usb 3-1: Manufacturer: FTDI
[48941.053382] usb 3-1: SerialNumber: FTHDEFGP
[48941.056929] ftdi_sio 3-1:1.0: FTDI USB Serial Device converter detected
[48941.057011] usb 3-1: Detected FT232RL
[48941.057016] usb 3-1: Number of endpoints 2
[48941.057020] usb 3-1: Endpoint 1 MaxPacketSize 64
[48941.057024] usb 3-1: Endpoint 2 MaxPacketSize 64
[48941.057028] usb 3-1: Setting MaxPacketSize 64
[48941.057393] usb 3-1: FTDI USB Serial Device converter now attached to ttyUSB1

reconbot:
Glad it works!

PR #413, opened 2014-11-07
Solves for issue #401

voodootikigod:
Parity handling with the new NAN string handling thanks to @sguilly.

### DIFF
--- a/src/serialport.cpp
+++ b/src/serialport.cpp
@@ -387,31 +387,27 @@ void EIO_AfterDrain(uv_work_t* req) {
   delete req;
 }
 
+// Change request for ticket #401 - credit to @sguilly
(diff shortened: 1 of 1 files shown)


PiyushRamuka:
I am having similar issue with PI3 Serial Port (ttyS0). The code works without any issues with none parity. It also works with "Even" parity on USB serial port. However, the communication fails with even parity. I don't have any option but to use even parity as the STM32F requires even parity for bootloader. 

I have modified the serialport.cpp as recommend by @voodootikigod  and @sguilly but no luck. Do I need to rebuild after changing the .cpp file? If so, how do i do this? Sorry but this is very new to me.

Please let me know how can i resolve this issue. Thanks in advance!

PR #437, opened 2015-01-09
fix serialport.list missing ports when ports have same id

heqichen:


voodootikigod:
@rwaldron ping

voodootikigod:
fixed in #487 which allows the operator to select how to handle this.

PR #438, opened 2015-01-13
Removed console.log's

ecksun:
Logging should be part of the library user, not the library itself.

Perhaps its a good idea to use the debug library instead?

### DIFF
--- a/serialport.js
+++ b/serialport.js
@@ -397,13 +397,13 @@ function SerialPortFactory() {
     try {
       factory.SerialPortBinding.close(fd, function (err) {
         if (err) {
-          console.log('Disconnect completed with error:' + err);
+          // console.log('Disconnect completed with error:' + err);
         } else {
-          console.log('Disconnect completed');
+          // console.log('Disconnect completed');
         }
       });
     } catch (e) {
-      console.log('Disconnect failed with exception', e);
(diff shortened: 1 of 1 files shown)


ecksun:
How about the debug library? Its a quite nice library, by default nothing is logged, but by using the DEBUG environment variable you can specify what you want to have logged. For example when using mocha if I want more debugging I can simply run:
    DEBUG=mocha mocha
Which will give me all debug information related to mocha, likewise I could do DEBUG=* to get all debug output

voodootikigod:
Love the suggestion to use debug, will integrate it shortly (next release) for now hte console.log was my mistake. :name_badge:

voodootikigod:
We do emit or return the callback.. this just provides more contextual data
as needed.

*Chris Williams*

@voodootikigod <http://twitter.com/voodootikigod> | GitHub
<http://github.com/voodootikigod>

*The things I make that you should check out:*
SaferAging <http://www.saferaging.com/> | JSConf <http://jsconf.com/> |
RobotsConf <http://robotsconf.com/> | RobotsWeekly
<http://robotsweekly.com/>

Help me end the negativity on the internet, share this
<http://jsconf.eu/2011/an_end_to_negativity.html>.

PR #448, opened 2015-01-21
Ensure we call the open callback asynchronously

### LINKED ISSUE #444: When using openImmediately false and a callback, callback is not called
In the docs and tests this case clearly never thought of. I actually need it for a new implementation I'm working on. Just making sure no one has a problem if I fix this. Either that or we need to change the docs.

from docs:
openImmediately (optional)

Attempts to open a connection to the serial port on process.nextTick. The default is true. Set to false to manually call open() at a later time.

callback (optional)

Called when a connection has been opened. The callback should be a function that looks like: function (error) { ... }

reconbot:
- previously called it synchronously on options errors
 - previously only called asynchronously on open errors
 - update docs to provide more info about the callback

Closes #444

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ Attempts to open a connection to the serial port on `process.nextTick`. The defa
 
 **_callback (optional)_**
 
-Called when a connection has been opened. The callback should be a function that looks like: `function (error) { ... }`
+Called asynchronously when a connection has been opened or there is an error during opening. The callback should be a function that looks like: `function (error) { ... }`
 
 ### .open (callback)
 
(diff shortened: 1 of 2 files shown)


phated:
This is an easy fix, but completely breaking change.  Constructors shouldn't be async in JS

phated:
Constructors should throw on failure to construct and openImmediately shouldn't be an option.  As stated in the Streams2 PR, you should just call open on a constructed serial port.  An extra function (`create` or something) could be exposed with that behavior.

phated:
That PR is too large for its own good and each of the ideas in it can (and should) stand alone.  You are breaking current behavior with this that people currently depend on.  If you are bumping major semver, you should fix the larger problem, not make the async stuff worse.

phated:
This is behavior that we rely on by setting openImmediately to false and treating the constructor synchronous.

Issue #426, opened 2014-12-15
There is only one combined write queue for all serial ports.


giseburt:
For the TinyG project (and those that use TinyG) we need multiple serial ports open, each with their own write queue. We "flood" one with writes of motion commands, and the other we use to send immediate commands like "pause" and "resume."

I have modified the code to have multiple write queues and write threads, but it needs more testing.

It is in the master branch here: https://github.com/giseburt/node-serialport

In particular I would appreciate it if someone could test Linux and Windows. It's working on OS X (Mavericks - 10.9.4).

I'm also concerned about the close procedure not releasing data properly if the port is closed with a large amount of data in the queue.

-Rob

reconbot:
This looks like it was solved!

PR #462, opened 2015-02-19
Fix for #426 -- multiple serial ports each get their own write queue (and thread)

giseburt:
This is the pull request mentioned in #426, merged and rebase with the current master.

Note that the diff shows that the entire file changed wrongly, due to mixed line endings being normalized. If you do a diff -b on the command line you'll see the changes are relatively minor and removes a fair bit of complexity.

I will later have another pull request that normalizes line endings, but that makes most of the files in the project appear to be replaced, and I didn't want to confuse things.

-Rob

### DIFF
--- a/src/serialport.cpp
+++ b/src/serialport.cpp
@@ -326,14 +326,16 @@ void EIO_AfterClose(uv_work_t* req) {
 
     _WriteQueue *q = qForFD(data->fd);
     if (q) {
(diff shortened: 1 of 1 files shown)


joshperry:
@giseburt Thank you for tackling this, we are using multiple serial ports as well and noticed this issue in the native code.

PR #470, opened 2015-03-02
Controlsignals

jacobrosenthal:
* Add windows control signaling. 
* Fix RTS and DTR example, they're default high so to do a reset you actually set them false. (this is consistent with Chrome's serial handling). 
* Fix RTS and DTR defaults, and since false is significant alter the default flush in the set function.
* Add new control signal called Break which sets TX high. Added for both windows and unix/linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,7 +317,7 @@ An object with the following properties:
 
 **_openImmediately (optional)_**
 
-Attempts to open a connection to the serial port on `process.nextTick`. The default is `true`. Set to `false` to manually call `open()` at a later time.
+Attempts to open a connection to the serial port on `process.nextTick`. The default is `true`. Set to `false` to manually call `open()` at a later time, but note you'll need to use factory error listener in the case of constructor errors. 
 
 **_callback (optional)_**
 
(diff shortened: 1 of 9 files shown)


jacobrosenthal:
Heh. I think well need to change our jshint rules.

phated:
:+1: using these APIs from the fork but would love them upstreamed

voodootikigod:
Merging tonight

voodootikigod:
first time trying to recompile with 0.12... grrr

*Chris Williams*

@voodootikigod <http://twitter.com/voodootikigod> | GitHub
<http://github.com/voodootikigod>

*The things I make that you should check out:*
SaferAging <http://www.saferaging.com/> | JSConf <http://jsconf.com/> |
RobotsConf <http://robotsconf.com/> | RobotsWeekly
<http://robotsweekly.com/>

Help me end the negativity on the internet, share this
<http://jsconf.eu/2011/an_end_to_negativity.html>.

phated:
Node 0.12 isn't a thing. `nave usemain 0.10.36`

PR #476, opened 2015-03-06
Add BREAK command, useful for CISCO router to break into rommon

jerome-42:
Hi,

I work with Cisco routers. When I would like to break the boot process and switch to rommon mode, I need to send BREAK signal on the serial line. This command is missing in node-serialport, so I added it.

Kind Regards,

### DIFF
--- a/serialport.js
+++ b/serialport.js
@@ -663,6 +663,35 @@ function SerialPortFactory() {
     });
   };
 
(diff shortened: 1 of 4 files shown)


jacobrosenthal:
Heh, Break was in my commit you added yesterday :)

PR #487, opened 2015-03-17
Configurable factory to query the list of devices.

voodootikigod:
Can now query serialports by-id or by-name addressing the issue #437


jacobrosenthal:
Looks like this failure is related to the revert #476 we're working on. Lets put it on top of #490 and try again

Issue #468, opened 2015-03-01
failed to compile in macos


jemos:
Hi,

It was working a few weeks back, but now I get:

...
> node-pre-gyp install --fallback-to-build

child_process: customFds option is deprecated, use stdio instead.
child_process: customFds option is deprecated, use stdio instead.
  CXX(target) Release/obj.target/serialport/src/serialport.o
  CXX(target) Release/obj.target/serialport/src/serialport_unix.o
../src/serialport_unix.cpp: In function ‘void EIO_Open(uv_work_t*)’:
../src/serialport_unix.cpp:146: error: ‘O_CLOEXEC’ was not declared in this scope
../src/serialport_unix.cpp: In function ‘void ExtractUsbInformation(stSerialDevice*, IOUSBDeviceInterface**)’:
../src/serialport_unix.cpp:456: warning: format ‘%08x’ expects type ‘unsigned int’, but argument 4 has type ‘UInt32’
make: *** [Release/obj.target/serialport/src/serialport_unix.o] Error 1
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:267:23)
gyp ERR! stack     at ChildProcess.emit (events.js:110:17)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (child_process.js:1067:12)
gyp ERR! System Darwin 10.8.0

Any idea?

Thank you.
Jean Mousinho

jacobrosenthal:
Hi Jean,

Can you get really specific on what you're trying to do? Is this just an npm i? What version node, node-serialport. Is the code on github you can link to so we can try?

Thanks,
Jacob

jemos:
Hi Jacob,

Sorry for the lack of information.

I can provide more information when I get home, but I can add the following
for now:

- When installing it will try to compile the code (is this a problem?) as
you probably noticed, but the source is using O_CLOEXEC constant which does
not exist in MacOS headers (Snow Leopard?).

- Trying with P_CLOEXEC makes compiling work, so module installs
successfully afterwards.

- Not sure if its functional since first experiment with a bluetooth serial
device did not work. Need to test with other device.


Thanks.

Regards,
Jean Mousinho

On Mon, Mar 2, 2015 at 8:08 AM, Jacob Rosenthal <notifications@github.com>
wrote:

> Hi Jean,
>
[...]

PR #499, opened 2015-04-10
set cloexec after open, possible fix for #468

jacobrosenthal:
@jemos can you verify


jemos:
Yes, it compiles now at least. Thanks.

PR #505, opened 2015-04-12
Refactor to multiple files

jacobrosenthal:
The current setup, with several scopes, is pretty rough to read and debug at this point. As a first step I wanted to see what it would look like to split out to separate files with basically no code changes. The big problem was the factory global which I'm now using a singleton to replicate.  I don't have all the tests passing, but tested open and read actual hardware so its mostly working. Wanted to get a discussion before I did the last 10%


voodootikigod:
@jacobrosenthal we should chat about this in person at jsconf. cool?

jacobrosenthal:
Thats fine with me. Im just hoping for a larger discussion with all contribs about about how to get more people taking ownership of smaller pieces.

PR #550, opened 2015-07-08
Removing BuildCommDCB for support on Windows 10 IoT Core

munyirik:
This change enables serialport to be used on devices running Windows 10 IoT Core (see http://ms-iot.github.io/content/en-US/GetStarted.htm). It’s a minor change to remove the call to BuildCommDCB which isn’t supported in Windows 10 IoT Core. Instead the DCB structure is populated 'manually' with the arguments that were passed to BuildCommDCB.
This is part of the work being done to bring Node.js to Windows 10 IoT Core. Background information can be found here - http://blogs.windows.com/buildingapps/2015/05/12/bringing-node-js-to-windows-10-iot-core/. [**Edit:** Previous link pointed to the wrong page]


src/serialport_win.cpp
```diff
@@ -82,10 +82,18 @@ void EIO_Open(uv_work_t* req) {
   dcb.DCBlength = sizeof(DCB);
   if(data->hupcl == false)
       dcb.fDtrControl = DTR_CONTROL_DISABLE; // disable DTR to avoid reset
-  if(!BuildCommDCB("9600,n,8,1", &dcb)) {
-    ErrorCodeToString("BuildCommDCB", GetLastError(), data->errorString);
-    return;
-  }
+  else
```
reconbot: As a nitpick, could you add `{}` around the if/else blocks?

munyirik: will do.

munyirik:
Thanks @reconbot. I'm also waiting on a response to the thread below. I'd like to know how I can help with setting up hardware at Microsoft for CI - https://github.com/nodejs/hardware/issues/13#issuecomment-129606036

mattpodwysocki:
@munyirik yes, I've been in talks with Dan and Arunesh on this and I think we are having some progress

PR #554, opened 2015-07-16
Add a simple start/end string delimiter parser and its respective test

lutangar:
Hey there,

I was wondering if this simple string parser could be of any use to you.
I used it to gather well formed json periodically broadcasted from an arduino micro.
```
var serialport = new SerialPort(path, {
    parser: parsers.stringDelimiters('{', '}')
});
```

Cheers.

### DIFF
--- a/parsers.js
+++ b/parsers.js
@@ -61,5 +61,19 @@ module.exports = {
         }
       }
     };
+  },
+
+  //encoding: ascii utf8 utf16le ucs2 base64 binary hex
+  stringDelimiters: function (startDelimiter, endDelimiter, encoding) {
+    encoding = encoding || 'utf8';
+
+    var data = '', rawData = '';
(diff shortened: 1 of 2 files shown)


parsers.js
```diff
@@ -61,5 +61,19 @@ module.exports = {
...
+  },
+
+  //encoding: ascii utf8 utf16le ucs2 base64 binary hex
+  stringDelimiters: function (startDelimiter, endDelimiter, encoding) {
+    encoding = encoding || 'utf8';
+
+    var data = '', rawData = '';
+    return function (emitter, buffer) {
+      rawData += buffer.toString(encoding);
+      if (rawData.indexOf(startDelimiter) >= 0 && rawData.indexOf(endDelimiter) >= 0) {
```
jupe: you are not validate here that startDelimiter is before endDelimiter

PR #561, opened 2015-08-10
Always query devices by name so hardware with duplicate IDs can be found

abrinsmead:


jacobrosenthal:
Feeling pretty good about this. We're having lots of trouble with by-id, especialy on arm systems. systemd doesnt appear to create it until a device is plugged in for instance on my novena (arm) I get a crash on listing ports until something is plugged in.

Need to get more platforms testing to see if theres anything were not thinking of before the switch

jacobrosenthal:
@abrinsmead what did you develop and test on so far?

PR #566, opened 2015-08-22
Upgrade to Nan 2.0.0

ghostoy:
Upgrade to Nan 2.0.0 to support latest io.js 3.x


ghostoy:
Sorry. I don't know how to make Travis-CI work correctly. But it's built and tested ok on my Windows, Linux and Mac.

zzz6519003:
:beer:  the commits can be squashed XD

rwaldron:
The failure is io.js 3.0.0 specific—maybe @rvagg can take a look and offer some insight.

rvagg:
this is a compiler incompatibility, Travis runs on Ubuntu 12.04 and therefore has a GCC that doesn't support the newer C++ functionality being embraced by V8, so you need to upgrade it (tbh I don't know about clang and I see you have `compiler: clang` in there but I'm guessing it's similar). What other projects have done to make this work is add the following:

```
addons:
  apt:
    sources:
    - ubuntu-toolchain-r-test
    packages:
    - g++-4.8
```

Then under `install` add the following at the top of the list:

```
  - if [[ $TRAVIS_OS_NAME == "linux" ]]; then export CXX=g++-4.8; fi
  - $CXX --version
```

(the `"linux"` case there is helpful if you're adding the experimental osx support travis have made available).

On your builds you should see a gcc version of 4.8. Probably remove the `clang` bit while you're there.

rwaldron:
Thanks @rvagg

src/win/stdstring.h
```diff
@@ -63,7 +63,7 @@
 //			- Jim Cline
 //			- Jeff Kohn
 //			- Todd Heckel
-//			- Ullrich Poll�hne
+//			- Ullrich Poll�hne
```
rwaldron: ?

ChALkeR:
Adding to the list: https://github.com/nodejs/node/issues/2798.

ghostoy:
Disabled 32bit build and dropped 0.8 because node-pre-gyp doesn't support 0.8 anymore.

ghostoy:
Bring back x86 builds on Travis-CI.

src/serialport.cpp
```diff
@@ -256,39 +254,38 @@ void EIO_AfterUpdate(uv_work_t* req) {
 }
 
 NAN_METHOD(Write) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

src/serialport.cpp
```diff
@@ -87,75 +87,74 @@ static void deleteQForFD(const int fd) {
...
+    Nan::ThrowTypeError("Third argument must be a function");
+    return;
   }
-  v8::Local<v8::Function> callback = args[2].As<v8::Function>();
+  v8::Local<v8::Function> callback = info[2].As<v8::Function>();
 
   OpenBaton* baton = new OpenBaton();
   memset(baton, 0, sizeof(OpenBaton));
   strcpy(baton->path, *path);
-  baton->baudRate = options->Get(NanNew<v8::String>("baudRate"))->ToInt32()->Int32Value();
```
jacobrosenthal: why remove all ->Get?

.travis.yml
```diff
@@ -1,44 +1,78 @@
...
+    packages:
+    - g++-4.8
+    - g++-4.8-multilib
+    - gcc-multilib
 env:
+  matrix:
+  # - TRAVIS_NODE_VERSION="0.8"
+  - TRAVIS_NODE_VERSION="0.10"
+  - TRAVIS_NODE_VERSION="0.12"
+  - TRAVIS_NODE_VERSION="iojs-3"
```
jacobrosenthal: why test io3?

.travis.yml
```diff
@@ -46,29 +80,26 @@ before_script:
...
-- export PATH=$(pwd)/node-${NVER}-${platform}-x86/bin:$PATH
-# install 32 bit compiler toolchain and X11
-- if [[ "$platform" == 'linux' ]]; then sudo apt-get -y install gcc-multilib g++-multilib; fi
-# test source compile in 32 bit mode with internal libsqlite3
-- if [[ "$platform" == 'linux' ]]; then CC=gcc-4.6 CXX=g++-4.6 npm install --build-from-source; else npm install --build-from-source; fi
-- npm test
-# publish 32 bit build
-- echo "Publishing x86 32bit Binary Package? ->" $PUBLISH_BINARY
-- if [[ $PUBLISH_BINARY == true ]]; then node-pre-gyp package publish; fi;
+# Disable 32bit testing temporily
```
jacobrosenthal: ouch

src/serialport.cpp
```diff
@@ -361,51 +358,50 @@ void EIO_AfterWrite(uv_work_t* req) {
...
 
-  NanDisposePersistent(data->buffer);
+  data->buffer.Reset();
   delete data->callback;
   delete data;
   delete queuedWrite;
 }
 
 NAN_METHOD(Close) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

src/serialport.cpp
```diff
@@ -432,50 +428,49 @@ void EIO_AfterClose(uv_work_t* req) {
 }
 
 NAN_METHOD(List) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

src/serialport.cpp
```diff
@@ -489,47 +484,46 @@ void EIO_AfterList(uv_work_t* req) {
 }
 
 NAN_METHOD(Flush) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

src/serialport.cpp
```diff
@@ -539,59 +533,58 @@ void EIO_AfterFlush(uv_work_t* req) {
 }
 
 NAN_METHOD(Set) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

src/serialport.cpp
```diff
@@ -601,47 +594,46 @@ void EIO_AfterSet(uv_work_t* req) {
 }
 
 NAN_METHOD(Drain) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

src/serialport_poller.cpp
```diff
@@ -50,55 +50,54 @@ void SerialportPoller::callCallback(int status) {
...
 
-  NanAssignPersistent<FunctionTemplate>(serialportpoller_constructor, tpl);
+  serialportpoller_constructor.Reset(tpl);
 
-  target->Set(NanNew<String>("SerialportPoller"), tpl->GetFunction());
+  Nan::Set(target, Nan::New<String>("SerialportPoller").ToLocalChecked(), Nan::GetFunction(tpl).ToLocalChecked());
 }
 
 NAN_METHOD(SerialportPoller::New) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

src/serialport_poller.cpp
```diff
@@ -111,20 +110,18 @@ void SerialportPoller::_stop() {
 
 
 NAN_METHOD(SerialportPoller::Start) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

src/serialport_poller.cpp
```diff
@@ -111,20 +110,18 @@ void SerialportPoller::_stop() {
...
 
-  SerialportPoller* obj = ObjectWrap::Unwrap<SerialportPoller>(args.This());
+  SerialportPoller* obj = Nan::ObjectWrap::Unwrap<SerialportPoller>(info.This());
   obj->_start();
   
-  NanReturnUndefined();
+  return;
 }
 NAN_METHOD(SerialportPoller::Close) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

src/serialport.cpp
```diff
@@ -87,75 +87,74 @@ static void deleteQForFD(const int fd) {
 
 
 NAN_METHOD(Open) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

src/serialport.cpp
```diff
@@ -172,74 +171,73 @@ void EIO_AfterOpen(uv_work_t* req) {
 }
 
 NAN_METHOD(Update) {
-  NanScope();
```
jacobrosenthal: Scope not added back? is it needed?

jacobrosenthal: From https://github.com/nodejs/nan/blob/master/doc/scopes.md

Allocate a new Nan::HandleScope whenever you are creating new V8 JavaScript objects. Note that an implicit HandleScope is created for you on JavaScript-accessible methods so you do not need to insert one yourself.


// new object is created, it needs a new scope:
void Pointless() {
  Nan::HandleScope scope;
  v8::Local<v8::Object> obj = Nan::New<v8::Object>();
}

// JavaScript-accessible method already has a HandleScope
NAN_METHOD(Pointless2) {
  v8::Local<v8::Object> obj = Nan::New<v8::Object>();
}


Seems like we would need these missing Nan::HandleScope scope; ?

rvagg: No, anything declared with `NAN_METHOD()` (or other `NAN_*()`) doesn't need it, that's the point of the text you copy and pasted above.

jacobrosenthal: found this list of exclusions here

https://github.com/nodejs/nan/blob/4ed41d760313b648b4d212d6ff5374668757be4f/tools/1to2.js#L75

.travis.yml
ghostoy: I copied this part from Nan, which tested against iojs-3. If you don't like, I can remove it.

.travis.yml
ghostoy: Forget to remove the comments.

src/serialport.cpp
ghostoy: It's not removed. It's replace with `Nan::Get`, which is said to be [compatible with all V8 versions](https://github.com/nodejs/nan/blob/master/doc/maybe_types.md#nanget).

jacobrosenthal: The double ToLocalChecked seems redundant and expensive, are you sure its needed?

ghostoy: `ToLocalChecked` is introduced by the new `Maybe` types, which fails at the point of null value. And there is another counterpart `ToLocal` which doesn't do the check. I think the current convention is assuming `options->Get` always return non-NULL value. So `ToLocalChecked` is the safest replacement.

imyller:
:+1: Travis CI passed. Good to go?

mgcrea:
Looking forward to have this merged to use with latest node stable.

fivdi:
I performed some tests with this PR on the BeagleBone Black with two different test setups.

1) Debian Wheezy, gcc/g++ 4.6, Node.js v0.10.38
   Result: installed successfully and a simple test ran successfully

2) Debian Jessie, gcc/g++ 4.9, Node.js v4.1.1
   Result: installed successfully and a simple test ran successfully

src/serialport.cpp
```diff
@@ -87,75 +87,74 @@ static void deleteQForFD(const int fd) {
...
-  baton->parity = ToParityEnum(options->Get(NanNew<v8::String>("parity"))->ToString());
-  baton->stopBits = ToStopBitEnum(options->Get(NanNew<v8::String>("stopBits"))->ToNumber()->NumberValue());
-  baton->rtscts = options->Get(NanNew<v8::String>("rtscts"))->ToBoolean()->BooleanValue();
-  baton->xon = options->Get(NanNew<v8::String>("xon"))->ToBoolean()->BooleanValue();
-  baton->xoff = options->Get(NanNew<v8::String>("xoff"))->ToBoolean()->BooleanValue();
-  baton->xany = options->Get(NanNew<v8::String>("xany"))->ToBoolean()->BooleanValue();
-  baton->hupcl = options->Get(NanNew<v8::String>("hupcl"))->ToBoolean()->BooleanValue();
-
-  v8::Local<v8::Object> platformOptions = options->Get(NanNew<v8::String>("platformOptions"))->ToObject();
+  baton->baudRate = Nan::Get(options, Nan::New<v8::String>("baudRate").ToLocalChecked()).ToLocalChecked()->ToInt32()->Int32Value();
```
jacobrosenthal: @ghostoy talking about this double tolocalchecked. Are we sure it needs to be checked twice?

rvagg: It's not being, the `"baudRate"` `v8::String` is being checked first and then the `v8::Value` that results from the `Nan::Get()` has to be converted too.

See https://nodesource.com/blog/cpp-addons-for-nodejs-v4 for some details on `MaybeLocal` types and why this garbage is needed (in short: blame Haskell).

voodootikigod:
This is missing the infrastructure update/buildout for Windows. That needs to be done in the appveyor.yml file. Will see if I can handle the update.

jacobrosenthal:
Squashed, rebased and commited as 866e4153b21e628335a7b9ef2dca471664dfb0c6

Issue #575, opened 2015-09-10
Allow setting `spfOptions`


lfk:
In my application I wish to enumerate two identical devices, which to my understanding wouldn't be a problem if I could explicitly set `spfOptions.queryPortsByPath` to `true`. Since the module is exported with an already instantiated factory, however, I've got way to actually set this option. As a matter of fact, it seems like the value won't ever be set to `true`.

Have I just missed something here?

PR #576, opened 2015-09-10
Expose `SerialPortFactory` method

lfk:
Addresses issue #575 

Allows the user to manually instantiate a new `SerialPortFactory`,
including an options object. Does not break backwards-compatibility.

Usage example:

```js
var serial = require('serialport'),
    serial = new serial.SerialPortFactory({ 'queryPortsByPath': true });
```

Querying ports by path is desireable when there is a need to interact
with multiple identical serial devices.


lfk:
Any feedback on this?

PR #580, opened 2015-09-12
Updated Raspberry Pi install instructions

nebrius:
Reworked the links to point to the pre-built deb packages that we recommend people use for raspi-io and johnny-five.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,17 +126,11 @@ npm install serialport
  * Download and install node.js:
 
 ```bash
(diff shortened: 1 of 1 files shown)


jacobrosenthal:
Heh, well that was hasty apparently. latest points at 4.0

https://node-arm.herokuapp.com/
NOTE: You will have to upgrade to jessie version of debian in order to run latest node!
Replace all instances of "wheezy" in /etc/apt/sources.list

Not sure thatll work @nebrius

PR #581, opened 2015-09-13
Triggering garbage collection less frequently

gregfriedland:
My project is writing to an Arduino over the serial port 50 times per second, which is causing garbage collection to get triggered every 30 seconds or so when running on a Raspberry Pi. This ends up locking up the RPi for the duration of gc.

node-serialport.read() and .write() use inline functions in the standard js way, which unfortunately reallocates their memory every time. See "1) Pull inner functions...into longer lived scope" in http://stackoverflow.com/a/18411275.

In my example, after this change memory allocation on the heap went from ~150kB over 10s to ~10kB.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,15 +54,16 @@
     "node-pre-gyp"
   ],
   "devDependencies": {
-    "mocha": "*",
-    "chai": "*",
-    "sinon-chai": "*",
-    "sinon": "*",
+    "chai": "^3.2.0",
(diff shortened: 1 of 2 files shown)


gregfriedland:
Afraid I'm having trouble with the tests. Any idea how to resolve this?
`TypeError: Cannot read property 'DEBUG_FD' of undefined`

PR #584, opened 2015-09-15
Update README for Pi/BBB/ARM users

keyvanfatehi:
Removed instruction to install "node-latest" via node-arm heroku site and instead install iojs 2.5.0 (which compiles correctly, unlike node 4, and does not leak memory in node-serialport code paths, unlike node 0.12.6) using wget & tar method directly from the official site.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,10 @@ sudo apt-get install build-essential
 npm install serialport
 ```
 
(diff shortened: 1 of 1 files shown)


jacobrosenthal:
closing for discussion at https://github.com/voodootikigod/node-serialport/issues/598

Issue #472, opened 2015-03-02
this.fd = 0 treated as an error


ecksun:
I just noticed that when the file descriptor returned from SerialPortBinding.open is 0 it is considered an error.

In most places the fd is checked with truthines, such as for example in the drain method:

```javascript
var fd = this.fd;
if (!fd) {
    // ...
}
```

In afterRead for example a check is made that `self.fd >= 0`, which suggest that 0 is in fact a valid file descriptor.

So my first question then is if 0 is a valid file descriptor?

In my experiments its seems to be, as I get a file descriptor of 0 from SerialPortBinding.open and if I fix all the checks I can communicate with my serial devices.

Secondly I think the checks for `self.fd >= 0` are broken, as if self.fd is null the expression will evaluate to true.

If I have correctly understood the problems I will submit a pull-request fixing the issues

reconbot:
It's my understanding that fd0 will always be stdin even if you've closed it. I support this change, but I'm curious has this caused a bug for you? And if so can you reproduce it?

ecksun:
Yes, my system does not work without this patch. I could reproduce it rather reliably but I never tried to reduce it to an easy example. It was a long time ago I found and fixed it so I don't remember much about how I found it.

Afaik the fd you get is defined by the kernel. Are you looking for a unit-test to show that this works? A proper system-test would probably require us to mess with the kernel or to mock the open system call

reconbot:
I pushed your changes in #703 , if you can still test this situation, I'd love to have you try it. (What operating system and device was it?)

ecksun:
I will see if I can get some time. This was several ARM systems running debian stable

reconbot:
You'll wan to test `serialport@2.0.7-beta2` on #688

ecksun:
Runnig 2.0.6 gives me this:

    nodejs: ../deps/uv/src/unix/core.c:701: uv__io_stop: Assertion `loop->watchers[w->fd] == w' failed.

While 2.0.7-beta2 works fine.

reconbot:
Whoop!

Thanks for fixing this! There will be at least one more beta. I cannot wait for this release.

PR #586, opened 2015-09-16
Bugfix/file descriptor 0

ecksun:
This is a patch that fixes the issue where file descriptor 0 is considered invalid (issue #472).

The tests does not explicitly define the fd to be 0, instead I use an expect in the test itself. If you feel this is inadequate I can change the tests to stub the open call to always return fd 0 for those tests.

### DIFF
--- a/serialport.js
+++ b/serialport.js
@@ -249,7 +249,7 @@ function SerialPortFactory(_spfOptions) {
   //verification code above
   SerialPort.prototype.update = function (options, callback) {
     var self = this;
-    if (!this.fd) {
+    if (this.fd === null) {
       debug('Update attempted, but serialport not available - FD is not set');
       var err = new Error('Serialport not open.');
       if (callback) {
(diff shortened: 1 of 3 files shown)


serialport.js
```diff
@@ -278,12 +278,12 @@ function SerialPortFactory(_spfOptions) {
   };
 
   SerialPort.prototype.isOpen = function() {
-    return (this.fd ? true : false);
+    return (this.fd !== null ? true : false);
```
pabigot: IMO this would be more clear as:

    return this.fd !== null;

since the conversion to a boolean value is implicit.

Also using `!this.isOpen()` or `!self.isOpen()` instead of replicating the specific value check wherever openness is tested would be more robust.

ecksun: Yes, indeed, I agree. I only tried to minimize the patch.

I just rewrote that commit

PR #588, opened 2015-09-18
Fixed capitalization of "PopcornTime"

djchie:
Changed "Popcorntime" to "PopcornTime"


jacobrosenthal:
Popcorn Time website has it as separate words, so if anything it should be that. PopcornTime is the spelling comsolid/rc4pt-node uses all over the page that that link goes to so might as well just leave it as that for now. Thanks though.

PR #593, opened 2015-09-22
Check `callback` before calling it in no-error case.

rodovich:
It appears to be an optional parameter because we check it before calling it in every other case.


rodovich:
@reconbot I made the change you suggested, & also updated the `drain` and `flush` callbacks since they followed the same pattern.

The `update` and `close` callbacks follow a similarly convoluted pattern but they do some other work before calling `callback` in the no-error case, so I left them alone.

rodovich:
Squashed.

Issue #618, opened 2015-10-28
serialport is crashing node program without errors or warning


livesankp:
Hello I am using latest version of serial port v2.0.2.
Running node.js program which receives events from server and writes/read to serialport.
It was working fine on node v 0.12 until i upgraded node to 4.1.2 and serialport too.
My node program is working but suddenly it stops without any error or warning.
Couple of times i have noticed CONNRESET but i have a feeling that serialport is crashing somehow during read/write. Couple of times i was able to run without crash too. But it is mostly crashing.
I have tried "git://github.com/ghostoy/node-serialport.git#nan1to2" this too but same issue.

Are you aware of any such issue ? I am using windows x86 for testing.

nicozhang:
i meet the same problem

koralm:
Same, serialport 2.0.2 with node 4.2.1 crashes. When i start server without physically connected serialport (no transmision) everything is ok. But when i start transmission after 1-3s server is suddenly stop.

livesankp:
Yes you are right. It happens for me after 2 mins.
Able to caught this error using uncaughtException
process.on('uncaughtException', function(error) {
  console.error((new Date).toUTCString() + ' uncaughtException:', error.message)
  console.error(error.stack);
  process.exit(1)
})

Error is 

Fri, 30 Oct 2015 19:25:07 GMT uncaughtException: read ECONNRESET
Error: read ECONNRESET
    at exports._errnoException (util.js:849:11)
    at TCP.onread (net.js:544:26)
exited : 1

EmbeddedMan:
I have also found this exact same problem. I have "fixed" it by reverting to Node.js version 0.12.7 and serialport version 1.7.4.

I sure hope that this issue can be solved in the latest serialport code, as I'd much rather be using the latest Node.js versions.

livesankp:
Right. Not sure what is causing this. When i run serialport and jsdom, jquery separately it is working. But the moment i combine both of them it starts crashing. Reverting everything back to your suggestion was always working i tried out that before.

jacobrosenthal:
Can any of you still reproduce this with the newly released 2.0.4?

livesankp:
@jacobrosenthal i tried this.  Looks like this issue is fixed with node 4.2.2 and serialport 2.0.4
Will let you know if this happens again.

jonfaust:
[...]

PR #623, opened 2015-11-05
Fix heap corruption issue on windows

### LINKED ISSUE #617: Versions 2.x crashes with Windows 7 x64
Running my tool with serialport 1.7.4 with node 0.12.7 worked well. After upgrading node-serialport to 2.x, the application became instable, crashing very often (terminates without any notice, very randomly). After reverting to 1.7.4 everything was fine again. Same effect when running serialport with node 4 (only serialport 2.x).

No issues on Mac OS, only Windows is affected. As 1.7.4 is not node 4 compliant, this is a big issue for me.

kunalspathak:
In ```EIO_AfterWatchPort```, serialport creates a `v8::ArrayBuffer` using `Nan::NewBuffer() ` API which creates `v8::ArrayBuffer` with ` ArrayBufferCreationMode::kInternalized ` mode. This means that v8 engine takes the responsibility of freeing the external buffer passed to the `v8::ArrayBuffer`.
However ```EIO_AfterWatchPort``` deletes and frees the buffer at the end of method.
```C++
cleanup:
    delete data;
    delete req;
```
When ArrayBuffer object is garbage collected, it tries to free the external buffer passed to it, but since it is already freed, we get heap corruption and the stack looks pretty much like one shown in #617.
The repro for this is inconsistent as it depends on state of memory at the time of GC, but with Microsoft's [AppVerifier](https://www.microsoft.com/en-us/download/details.aspx?id=20028) tool, there is a consistent repro for this.

The fix is to pass `FinalizerCallback` to `Nan::NewBuffer()` API that will internally create the ArrayBuffer in `ArrayBufferCreationMode::kExternalized` mode. This means that engine won't take the responsibility of freeing the buffer and when collection of arraybuffer happens, it will call the callback which will free the buffer.

Fixes #617, #618

### DIFF
--- a/src/serialport_win.cpp
+++ b/src/serialport_win.cpp
@@ -290,6 +290,16 @@ void DisposeWatchPortCallbacks(WatchPortBaton* data) {
   delete data->disconnectedCallback;
 }
 
+// FinalizerCallback will prevent WatchPortBaton::buffer from getting
+// collected by gc while finalizing v8::ArrayBuffer. The buffer will
+// get cleaned up through this callback.
+static void FinalizerCallback(char* data, void* hint) {
+  uv_work_t* req = reinterpret_cast<uv_work_t*>(hint);
+  WatchPortBaton* wpb = static_cast<WatchPortBaton*>(req->data);
+  delete wpb;
+  delete req;
+}
+
 void EIO_AfterWatchPort(uv_work_t* req) {
   Nan::HandleScope scope;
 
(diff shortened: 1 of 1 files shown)


Khhroom:
Just checked out this changes. They actually fix the problem with crashing under Windows.

techninja:
Yep, just tested, this fixes a pretty nasty anonymous crash in Electron as well!

PR #631, opened 2015-11-23
Adds .npmignore

rwaldron:
These things don't need to be shipped to npm.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,14 @@
+arduinoTest/
+examples/
+hwtest/
+sandbox/
+test/
+test_mocks/
+.jshintrc
+.travis.yml
+AUTHORS
+appveyor.yml
+changelog.md
+Gruntfile.js
+publish-binaries.md
+


jacobrosenthal:
Agreed. Though we need to keep our eye on this. I got a security email from npm recently. Apparently npm uses the gitignore if npmignore isnt there, but ignores the gitignore if npmignore IS there. This meant our gitignored gyp build directory was being shipped, which was leaking our npm credentials. Thus I flattened our old npmignore back into gitignore
https://github.com/voodootikigod/node-serialport/commit/ac8e598f44f2d4499d0d3a24a696b0082b71ac11

So we can do this, but we need to mirror our gitignore into here as well. Perhaps a note in the both as well to remind us why.

rwaldron:
@jacobrosenthal sorry for the delay in getting back to you—should I just make sure that .gitignore entries are also included here? (Sorry if that seems dense ;)

jacobrosenthal:
Yes under a gitignore section with a comment to remember to keep them synced, perhaps pointing to this issue

jacobrosenthal:
Added in e27745c44bf0980af3f18cfb979963dff84a47ae for 2.0.6 release

PR #662, opened 2016-01-26
Remove deprecated BuildCommDCB for windows 10 support

reconbot:
- This replaces the removed shortcut methods to build a serial port
object with the non deprecated long form.
- Unit and Smoke tested on windows 10.
- Replaces #550 as that source tree has disappeared.

cc @munyirik @mattpodwysocki

Tested on
 - [x] Windows 7
 - [x] Windows 8.1
 - [x] Windows 10
 - [ ] Windows 10 IoT Core

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ This assumes you have everything on your system necessary to compile ANY native 
 
 #### Windows:
 
- * Windows 7 or Windows 8.1 are supported.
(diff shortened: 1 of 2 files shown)


mattpodwysocki:
@reconbot yes, that was the intent from the Hardware group in Node.js cc @rwaldron

jacobrosenthal:
done
`npm i serialport@2.0.7-beta1`

Issue #661, opened 2016-01-26
close problem with moxa products (file descriptor not close)


mathieuwlo:
Hello,

I'm using moxa product like Uport 1650 or nport 5650, there isn't any problem to open a serial port but when I call close() function of node-serialport it doesn't close the port (file descriptor) so I can't reopen it whitout relaunching my node application.
I'm working on ubuntu 14.04 with kernel 3.19, node version 0.12.7
node serial port version 2.0.6
when I launch my node application with this command
strace -e trace=open,close node myprogramm.js I saw the open("/dev/tty",O_RDWR,O_CLOEXEC)= 13 (number of file descriptor) but when the close function is called I never saw the trace closing the file descriptor open before.

thanks

gfcittolin:
I had a similar problem, and found the problem to be a double call to `open()`, generating two file descriptors, and losing track of the first one:

https://github.com/voodootikigod/node-serialport/blob/master/src/serialport_unix.cpp#L134 and
https://github.com/voodootikigod/node-serialport/blob/master/src/serialport_unix.cpp#L186

Easy to see with some debug printf's:
```
* JS call - port.open
>>>>>> EIO_Open - before open path:/dev/ttyACM0
>>>>>> EIO_Open - after open path:/dev/ttyACM0, fd:12       <-- from L134
>>>>>>>> setup - fd:12
>>>>>>>> setup - before open fd:12
>>>>>>>> setup - after open fd:13                           <-- from L186, fd 12 gets lost
>>>>>>>> setup - before fcntl fd:13
>>>>>>>> setup - after fcntl fd:13, cloexec:0
>>>>>>>> setup - before tcgetattr fd:13
>>>>>>>> setup - after tcgetattr fd:13
>>>>>>>> setup - before cfsetispeed fd:13, baudRate:13
>>>>>>>> setup - after cfsetispeed fd:13, baudRate:13
>>>>>>>> setup - before cfsetospeed fd:13, baudRate:13
>>>>>>>> setup - after cfsetospeed fd:13, baudRate:13
>>>>>>>> setup - before tcflush fd:13
>>>>>>>> setup - before tcsetattr fd:13
>>>>>>>> setup - after tcsetattr fd:13
* JS callback - onOpen
```

The `open("/dev/tty",O_RDWR,O_CLOEXEC)= 13 ` you are seeing is node opening the terminal, not your serial device

reconbot:
[...]

PR #670, opened 2016-02-16
Fixes double open() call on unix

gfcittolin:
Fixes a double call to open() on unix, that leaks a file descriptor and keeps the underlying device open, as discussed on #661

### DIFF
--- a/src/serialport_unix.cpp
+++ b/src/serialport_unix.cpp
@@ -131,6 +131,9 @@ void EIO_Open(uv_work_t* req) {
   OpenBaton* data = static_cast<OpenBaton*>(req->data);
 
   int flags = (O_RDWR | O_NOCTTY | O_NONBLOCK | O_CLOEXEC | O_SYNC);
+  if(data->hupcl == false) {
+    flags &= ~HUPCL;
+  }
   int fd = open(data->path, flags);
 
   if(-1 == setup(fd, data)){
@@ -171,20 +174,6 @@ int setup(int fd, OpenBaton *data) {
     snprintf(data->errorString, sizeof(data->errorString), "Invalid data bits setting %d", data->dataBits);
     return -1;
   }
-
-
-  // int flags = (O_RDWR | O_NOCTTY | O_NONBLOCK | O_CLOEXEC | O_SYNC);
-  // if(data->hupcl == false) {
-  //   flags &= ~HUPCL;
-  // }
-  // int fd = open(data->path, flags);
-
-  int flags = (O_RDWR | O_NOCTTY | O_NONBLOCK | O_CLOEXEC | O_SYNC);
-  if(data->hupcl == false) {
-    flags &= ~HUPCL;
-  }
-  fd = open(data->path, flags);
-
 
   if (fd == -1) {
     snprintf(data->errorString, sizeof(data->errorString), "Cannot open %s", data->path);


shoerob:
I'm encountering a problem on OS X that might be related to this. When I open() my device, the data comes through just fine. But if I close() it, then open() it again, no data comes through anymore. I have to unplug it, and plug it back in again in order to open() and receive data again. The strange thing is that open() doesn't complain when this is happening, just no data comes through.

JonathonAshworth:
I'm having the exact same problem as shoerob, managed to find it here through search but there doesn't seem to be an open issue. Is this merge really a fix for it? Seems unrelated... but hey, whatever works!

How long until this is accepted?

limbera:
I have tested this on my mac experiencing a similar sounding issue to @shoerob but it doesn't fix it for me.

OSX 10.11.3
Running Electron

limbera:
@shoerob 

Our application lists all serial ports and then sends some test packets to identify more information about the port. We found that the only working on unplug and replug issue was actually being caused by the '/dev/cu.Bluetooth-Incoming-Port'. When we filtered that port out of the list of ports which were sent test packets, the problem disappeared.

limbera:
I am experiencing a similar issue to @shoerob but this fork did not solve it for me.

JonathonAshworth:
The relevant code is quite chunky and we're having trouble narrowing down the problem. I've put it in a gist, but I'm going to try and come up with a minimum representation of the problem now so you've got something a bit cleaner to work with!

https://gist.github.com/JonathonAshworth/a099183982a8c40747ab

Essentially, our code opens each serial port on the computer in turn, sends a test packet to verify that the device responds to the MTK format, then requests firmware information from the device to verify that it's a valid device that was bought from our company. Then it closes the port.

This happens for every valid serial port on the computer, generating an array of "valid ports" (i.e ports which represent one of our devices). Then those valid ports are opened a second time so data can be downloaded off them. On the second open, a test packet is once again sent, then a packet requesting the memory status of the device is sent, followed by downloading the actual memory off the device.

The problem appears to be with the open-close-open process, and happens only on OSX. If we do not check the validity of a device first, then it downloads just fine, but if the device has been opened and closed previously, then it will be successfully reopened, but will fail later on. Half the time it fails to respond to the test packet, and half the time it successfully responds to the test packet, then fails to respond to the memory status request.

The curious part to me at least is that it's sometimes successfully responding to the test packet before failing. It seems as if the device is being randomly closed or randomly failing.

JonathonAshworth:
https://gist.github.com/JonathonAshworth/ecafc8a2df376a4063e8

Alright, this is the simplest test I could come up with that is failing. The device constantly outputs a stream of realtime gps data, so the test just opens the device for 5 seconds, logging it's output to console, closes it, then re-opens it for another 5 seconds.

This acts as you'd expect on both Ubuntu and Windows, but on Mac, it only ever outputs the very first time the device is opened after plugging it in. The second time it's opened nothing prints, and on subsequent runs it doesn't print either time, until the device is unplugged and plugged in again.

My workmate just installed this fork and is reporting that it's working as expected. So it appears this is (somehow) a fix for the issue.

gfcittolin:
@JonathonAshworth is your device a physical serial device, or it has
another interface (like usb) and just emulates a serial port?

This issue you have may be related to how the driver is reacting to the
"leaked" handle. I had exactly the same issue you're having, but with
Windows. A software in Labview was not properly closing a handle to our
device before exiting, and we were not able to read from it again before
unplugging and plugging it back.

We have an USB device that emulates a serial port for analog signal
acquisition, and we react to the "virtual RTS" signal. In linux, the USB
CDC driver will set this to 1 when it has one or more open fds to the
device, and then back to 0 when all are closed. In this case, it was on
even after closing, what made me look into this.

May be the case either the OS driver is having problems with this left-open
handle, or even your device, with signals like RTS or DTR.

src/serialport_unix.cpp
```diff
@@ -131,6 +131,9 @@ void EIO_Open(uv_work_t* req) {
   OpenBaton* data = static_cast<OpenBaton*>(req->data);
 
   int flags = (O_RDWR | O_NOCTTY | O_NONBLOCK | O_CLOEXEC | O_SYNC);
+  if(data->hupcl == false) {
```
pabigot: This is a good patch and I'll incorporate it into my fork but this block should be removed, not moved.  `HUPCL` is not an fcntl flag; it's a tty control flag and has nothing to do with open(2).  On Linux x86_64 HUPCL happens to have the same value as O_APPEND so is benign but on other platforms this code might change the intended open behavior.

pabigot: Actually it should be moved within the setup() function.

Issue #686, opened 2016-03-24
race condition closing serial port causes segfault


pabigot:
I'm using node-serialport on Linux under Node 4.4.0 on Ubuntu 14.04, in a highly asynchronous application involving ES6 promises.  I've encountered a situation where node aborts with core dump in uv__io_poll when epoll_ctl(2) returns EBADF because it was passed a descriptor for a closed serial port.

The hand-waving summary is that if SerialPort.close() is invoked while a read is active, the descriptor can be closed before a pending EAGAIN error is processed by afterRead.  This causes the poller to be restarted with a closed descriptor, which breaks libuv.

I'll submit a trivial patch that fixes the issue for me.

However, I'm also skeptical of the original code. The `fd` property is variously set to `null` or `0` when the port is closed, but the check for re-enabling the poller in that case is `self.fd >= 0`, so it's probably always going to pass.

I'd recommend consistently using `-1` as the value of the `fd` property when the port is closed as that value won't convert to something that looks like a valid descriptor.

reconbot:
This is a great catch, thank you.

reconbot:
fixed in serialport@2.0.7-beta2

reconbot:
I'm going to close this issue as I've pretty throughly tested this in linux on `2.0.7-beta4`. Thanks for raising and helping fix this issue.

PR #687, opened 2016-03-24
fix race condition closing serial port

pabigot:
If SerialPort.close() is invoked while fs.read() is blocked the
post-read callback can re-enable the poller after the descriptor has
been closed.  Add a check to prevent this.

Signed-off-by: Peter A. Bigot <pab@pabigot.com>

### DIFF
--- a/serialport.js
+++ b/serialport.js
@@ -342,7 +342,7 @@ function SerialPortFactory(_spfOptions) {
         self.reading = false;
         if (err) {
           if (err.code && err.code === 'EAGAIN') {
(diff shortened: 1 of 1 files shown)


pabigot:
References issue #686.

pabigot:
I added the fix to the concern I mentioned in #686 about inconsistent detection of closed ports.  The tests failed because the mock serial port uses the fd property to hold the string path for the mocked device.

I also now see this is related to issue #472 and PR #586 by @ecksun, and endorse that solution instead, with one comment added.

PR #694, opened 2016-03-27
Update the README and add an ISSUE_TEMPLATE

reconbot:
I've been doing a lot of issue triage lately and I have found that an issue template would help a ton.

I've also been trying to refer to our README over and over and an index would go a long way.  Furthermore often I don't know what the heck people are expecting in their callbacks and we don't always tell them, so I made sure to be a little more clear.


jacobrosenthal:
We should probably just give them the commands to run to get the versions we want. (output of node -v etc)
We could ask for the npm debug log, but Ive noticed especially with node-gyp its impossible to read/doesnt have anything in it. I wonder if theres something better or some way to tweak it to make it readable.

PR #703, opened 2016-03-28
Bugfixes around JS handling of file descriptors

reconbot:
This patch combines #586 and #686

A big thanks to @ecksun and @pabigot for discovering and helping make these changes. 

I started to cleanup our logic around checking if the port was open.

Disconnections are weird, but no weirder than before this patch. @ecksun's notes below.

 - A closed serialport now uses fd = null
 - Emulate a proper fd in mock hardware
 - Instead of using the device path as the file descriptor, use a generated
integer as the fd.

The open(2) manpage states that the file descriptor will be a small nonnegative integer. Checking if the fd is falsey leads to the incorrect classification of the fd 0 as invalid. That is not the case, open(2) specifies the fd as "a small, nonnegative integer".

This also adds tests for all the cases where the fd was previously incorrectly considered invalid, which required adding three more methods to the mock-hardware (update, set and drain).

This solves issue #472

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -9,7 +9,7 @@ module.exports = function(grunt) {
       }
     },
     eslint: {
-      src: ['*.js', 'test/**/*.js', 'arduinoTest/**/*.js']
+      src: ['*.js', 'test/**/*.js', 'test_mocks/**/*.js', 'arduinoTest/**/*.js']
     }
   });
 
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
   "license": "MIT",
   "scripts": {
     "install": "node-pre-gyp install --fallback-to-build",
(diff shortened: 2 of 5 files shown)


ecksun:
I would suggest having different commits for the different issues, the self -> this change is semantically different from the fd != null change. Having them in different commits makes the changes easier to review and the history easier to understand and navigate.

pabigot:
There are at least three changes here: changing how a closed serial port is identified, removing some (but not all) unnecessary aliasing `self` variables in that module, and fixing the race condition from #686.  I concur with @ecksun on the value of using separate commits instead of bundling everything together.

I pushed a couple minor fixes on top of this to [my fork](https://github.com/pabigot/node-serialport/tree/bugfix/file_descriptor_0).  That branch (with my changes, didn't test the original) passes the code that originally brought me to report #686.

PR #705, opened 2016-03-29
flush also output buffer (TCIOFLUSH)

jupe:
use TCIOFLUSH instead of TCIFLUSH on linux build
```
TCIFLUSH
flushes data received but not read.
TCIOFLUSH
flushes both data received but not read, and data written but not transmitted.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,7 +297,7 @@ Resumes a paused connection.
 
 ### .flush (callback)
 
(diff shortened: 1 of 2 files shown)


soundanalogous:
Does `TCIOFLUSH` dump the data in the TX buffer before it's transmitted? That's sorta what the description sounds like. Looking over the docs, `tcdrain()` actually seems to be more inline with common definitions for 'flush' (at least implementations I'm more familiar with) which is transmitting and clearing the TX buffer. I don't think there is any clear standard however for the behavior of flush.

rwaldron:
@soundanalogous thanks for taking a look and weighing in :)

soundanalogous:
according to the docs, `tcdrain()` the behavior appears to be equivalent to Arduino's `Stream.flush()` function.

PR #707, opened 2016-03-30
Upgrade to NAN v2.2.1 for Node.js v6.0.0-nightly compatability

fivdi:
node-serialport currently depends on NAN ~2.0.0 which is too old for Node.js v6.0.0-nightly20160307061ebb39c9. Installing node-serialport with Node.js v6.0.0-nightly20160307061ebb39c9 results in compile errors. Upgrading to NAN v2.2.1 solves the problem.

Note that this PR wasn't tested a lot.


fivdi:
Sorry, I appear to have messed that up. I need to learn how to use git. I'm going to close this PR as it now looks like junk.

fivdi:
reopened as #712

Issue #697, opened 2016-03-27
Turn down the verbose logging on appveyor builds


reconbot:
It's super verbose, I'm not sure where that gets set, I think it's in the app veyor config @jacobrosenthal

Issue #651, opened 2015-12-27
Add more tty* path support for listLinux()


dbaba:
I've found that `listLinux()` function discards paths under `/dev/` other than `ttyS*`, `ttyACM*`, and `ttyUSB*`.

https://github.com/voodootikigod/node-serialport/blob/master/serialport.js#L579

Intel Edison Yocto, however, offers some serial ports which don't match the above filtering rules, e.g. `/dev/ttyMFD0`.

I think the line can be replaced with the following.
```javascript
if (!files[i].match(/tty[a-zA-Z]/) || !fs.statSync(path.join(dirName,files[i])).isCharacterDevice()){
```
How about that?

dbaba:
@voodootikigod I found the similar issue #610 
Should I re-open the issue with concrete port names and close this?
I can list the all required file names.

reconbot:
@dbaba a full list would be great. I'm not up to speed on this issue but it seems like something easy to tackle. Yocto is something I'd like to support better. We should also probably evaluate using `fs.statSync` =(

reconbot:
This one is fixed too, if you have more port names let us know!

Issue #621, opened 2015-11-05
double prefix "0x" vendorId and productId on ubuntu 15.10, serialport=2.0.2


khomyakov42:
`{
"comName":"/dev/ttyUSB0",
"manufacturer":"FTDI",
"serialNumber":"FTDI_FT232R_USB_UART_A602TQJB",
"pnpId":"pci-0000:00:13.0-usb-0:1:1.0-port0",
"vendorId":"0x0x1002",
"productId":"0x0x4397"
}`

reconbot:
Fixed in master and will be in the next beta. =) Please it out and let us know if it works for you!

PR #708, opened 2016-03-30
Move listUnix to its own module

### LINKED ISSUE #650: Command failed error on listing ports (udevadm is missing)
Listing ports on Intel Edison Yocto Linux always fails.

Tested Version: 2.0.5/2.0.6

```
var serialp = require("serialport");
serialp.list(function (err, ports) { console.log(err,ports); });
```
This shows the following error message.
```
{ [Error: Command failed: /bin/sh: /sbin/udevadm: No such file or directory
/bin/sh: /sbin/udevadm: No such file or directory
] killed: false, code: 127, signal: null } [ undefined ]
```
However, udevadm does exist there:
```
$ which udevadm
/bin/udevadm
```

reconbot:
- Refactor and fix bugs 
 - Remove the path behind `udevadm` as it’s not always in the same place but it’s so far it is always in the path
 - Tests listUnix, (`pnpId` assumes a lot about `udevadm` =( )
 - Adds `ttyMDF` to the list of serial ports for “High Speed UART device of Intel Medfield platform”

Fixes #650, #651, #634, #697 and #621


lib/list-unix.js
```diff
@@ -0,0 +1,92 @@
...
+
+    // get only serial port names
+    // TODO statSync is bad bad bad
+    for (var i = files.length - 1; i >= 0; i--) {
+      if (
+          (files[i].indexOf('ttyS') === -1 &&
+            files[i].indexOf('ttyACM') === -1 &&
+            files[i].indexOf('ttyUSB') === -1 &&
+            files[i].indexOf('ttyAMA') === -1 &&
+            files[i].indexOf('ttyMFD') === -1) ||
```
rwaldron: `!(/tty(S|A|U)/g).test(files[i])`

lib/list-unix.js
```diff
@@ -0,0 +1,92 @@
...
+
+function udev_parser(udev_output) {
+  var as_json = udev_output_to_json(udev_output);
+
+  var pnpId;
+  if (as_json.devlinks) {
+    as_json.devlinks.split(' ').forEach(function(path) {
+      if (path.indexOf('/by-id/') === -1) { return }
+      pnpId = path.substring(path.lastIndexOf('/') + 1);
+    });
```
reconbot: This fixes a bug, there are two "paths" and one of them is `by-id` but they come in different orders sometimes, the `forEach` isn't strictly necessary I suppose

rwaldron: Got it.

test/list-unix.js
```diff
@@ -0,0 +1,100 @@
...
+    'id_model_id=0043\n' +
+    'DEVLINKS=/dev/serial/by-path/pci-0000:00:14.0-usb-0:2:1.0-port0 /dev/serial/by-id/pci-NATA_Siolynx2_C8T6VI1F-if00-port0\n',
+  'ttyAMA_im_a_programmer':
+    'DEVNAME=/dev/ttyAMA_im_a_programmer\n' +
+    'DEVLINKS=/dev/serial/by-id/pci-NATA_Siolynx2_C8T6VI1F-if00-port0 /dev/serial/by-path/pci-0000:00:14.0-usb-0:2:1.0-port0\n',
+  'ttyMFD0':
+    'DEVNAME=/dev/ttyMFD0\n' +
+    'id_vendor_id=0x2341\n' +
+    'id_model_id=0x0043\n'
+};
```
rwaldron: Great fixtures, but I wonder if you could put these in a file?

lib/list-unix.js
reconbot: I know which kernel modules load most of these dev paths, I'm unsure of how many we don't actually want.

test/list-unix.js
```diff
@@ -0,0 +1,100 @@
...
+    'id_vendor_id=2341\n' +
+    'id_model_id=0043\n' +
+    'DEVLINKS=/dev/serial/by-path/pci-0000:00:14.0-usb-0:2:1.0-port0 /dev/serial/by-id/pci-NATA_Siolynx2_C8T6VI1F-if00-port0\n',
+  'ttyAMA_im_a_programmer':
+    'DEVNAME=/dev/ttyAMA_im_a_programmer\n' +
+    'DEVLINKS=/dev/serial/by-id/pci-NATA_Siolynx2_C8T6VI1F-if00-port0 /dev/serial/by-path/pci-0000:00:14.0-usb-0:2:1.0-port0\n',
+  'ttyMFD0':
+    'DEVNAME=/dev/ttyMFD0\n' +
+    'id_vendor_id=0x2341\n' +
+    'id_model_id=0x0043\n'
```
rwaldron: Property keys will be uppercased

lib/list-unix.js
reconbot: Which is to say I'm happy to keep patching more known good dev paths until we have an answer to that. But I can clean it up with a regex.

test/list-unix.js
rwaldron: I can't find any examples of hex values for ID_VENDOR_ID and ID_MODEL_ID

test_mocks/list-unix.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+    child_process: {
+      exec: function(cmd, cb) {
+        var port = cmd.split(/\/dev\/(.*)\)/)[1];
+        process.nextTick(function() {
+          cb(null, mockPorts[port]);
+        });
+      }
+    }
+  }
```
rwaldron: This project is using sinon, so you can def benefit from using a sinon sandbox for lifecycle managed stubs and spies

lib/list-unix.js
reconbot: Nah doing it without an iterative function isn't easier to understand `Array#find` would be the right answer.

test_mocks/list-unix.js
reconbot: I agree. This injects at require time not during test time. It would be hard to stub them once the module is loaded.

lib/list-unix.js
```diff
@@ -0,0 +1,83 @@
...
+        return callback(null, []);
+      }
+      return callback(err);
+    }
+
+    // get only serial port names
+    // TODO statSync is bad bad bad
+    for (var i = files.length - 1; i >= 0; i--) {
+      if (
+          !(/tty(S|ACM|USB|AMA|MFD)/).test(files[i]) ||
```
rwaldron: That works too ;)

lib/list-unix.js
```diff
@@ -0,0 +1,83 @@
...
+  var udevInfo = {};
+  var lines = output.split('\n');
+  lines.forEach(function(line) {
+    if (!line || line.trim() === '') {
+      return;
+    }
+    var line_parts = line.trim().split('=');
+    var key = line_parts[0].trim().toLowerCase();
+    udevInfo[key] = line_parts[1].trim();
+  });
```
rwaldron: Well in thaaaaaaaaat case: 

```js
var udevInfo = output.split('\n').reduce(function(accum, line) {
  if (!line || line.trim() === '') {
    return accum;
  }
  var parts = line.trim().split('=').map(function(part) {
    return part.trim();
  });

  accum[parts[0].toLowerCase()] = parts[1];

  return accum;
}, {});
```

lib/list-unix.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  var vendorId = udevInfo.id_vendor_id;
+  if (vendorId && vendorId.substring(0, 2) !== '0x') {
+    vendorId = '0x' + vendorId;
+  }
+
+  var productId = udevInfo.id_model_id;
+  if (productId && productId.substring(0, 2) !== '0x') {
+    productId = '0x' + productId;
```
rwaldron: Still haven't seen any examples in all of the docs, tutorials, articles, SO questions, etc. that show these values having `0x` prefix

lib/list-unix.js
reconbot: ahh no kidding, that's in node 0.10?!

lib/list-unix.js
reconbot: #621 and https://github.com/spark/particle-cli/issues/190#issuecomment-171656487

lib/list-unix.js
rwaldron: reduce? Yes.

test_mocks/list-unix.js
rwaldron: Disagree. 

- https://github.com/tessel/t2-cli/blob/master/test/unit/deploy.js#L1709-L1724
- https://github.com/tessel/t2-cli/blob/master/test/unit/deploy.js#L1837
- https://github.com/tessel/t2-cli/blob/master/test/unit/deploy.js#L1900-L1918
- https://github.com/tessel/t2-cli/blob/master/test/unit/deploy.js#L1934-L1936
- https://github.com/tessel/t2-cli/blob/master/test/unit/deploy.js#L1944

And tons more

reconbot: Well, you learn something every day. I'm not up for changing it now. I don't think this approach harms anything.

lib/list-unix.js
rwaldron: Nice! Also, I remembered something and dug through Tessel 2 code and discovered the same there. 

Why do you think this value is a hex string and not just evaluated to the decimal value?

reconbot: Someone who came before figured that part out, I have no idea if it's true
but I don't want to change the behavior unless it's actually a bug. So far
nobody has filled that it is.

Issue #682, opened 2016-03-17
serialport.node: __strcpy_chk: symbol not found on alpine linux


ArtworkAD:
I am using alpine linux 3.3 and node 4.3.

```
> serialport@2.0.6 install /app/node_modules/serialport
> node-pre-gyp install --fallback-to-build

[serialport] Success: "/app/node_modules/serialport/build/Release/node-v46-linux-x64/serialport.node" is installed via remote
serialport@2.0.6 node_modules/serialport
├── bindings@1.2.1
├── async@0.9.0
├── sf@0.1.7
├── debug@2.2.0 (ms@0.7.1)
├── nan@2.0.9
└── optimist@0.6.1 (wordwrap@0.0.3, minimist@0.0.10)
```

When I try to open a port I get an error:

```
npm ERR! Linux 3.19.0-30-generic
npm ERR! argv "/usr/bin/node" "/usr/bin/npm" "start"
npm ERR! node v4.3.0
npm ERR! npm  v2.14.12
npm ERR! code ELIFECYCLE
npm ERR! app@1.2.0 start: `node ./bin/www`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the app@1.2.0 start script 'node ./bin/www'.
npm ERR! This is most likely a problem with the app package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node ./bin/www
npm ERR! You can get their info via:
npm ERR!     npm owner ls app
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /app/app/npm-debug.log
module.js:435
  return process.dlopen(module, path._makeLong(filename));
                 ^

Error: Error relocating /app/app/node_modules/serialport/build/Release/node-v46-linux-x64/serialport.node: __strcpy_chk: symbol not found
    at Error (native)
    at Object.Module._extensions..node (module.js:435:18)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/app/app/node_modules/serialport/serialport.js:14:25)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
```

Everything worked fine with debian jessi and node 4.3. After switching to alpine linux I get the above error.

[...]

PR #710, opened 2016-03-30
Add instructions for Alpine Linux

mhart:
Addresses #682

Alpine is a small distro, increasingly being used in Docker environments (see https://github.com/mhart/alpine-node for an example image that includes node)

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,21 @@ sudo apt-get install build-essential
 npm install serialport
 ```
 
+#### Alpine Linux:
+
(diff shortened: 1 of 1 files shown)


mhart:
Ah, good question – I got that from https://github.com/voodootikigod/node-serialport/issues/433#issuecomment-68492099

Lemme try without.

On a related note – is there a way for serialport to just recognise that it can't use the pre-compiled lib in a musl environment, so the `--build-from-source` flag wouldn't even be necessary?

mhart:
Hmmm, so specifying it that way is "safer" if you've got other packages that you're also installing: https://github.com/mapbox/node-pre-gyp#options

In the example I have in the README, that's not the case, so it would work without, but in other install contexts that's not necessarily true.

Happy with whatever you prefer?

PR #717, opened 2016-04-02
Free the function from the factory

reconbot:
No longer make use of the serial port factory

This is a tiny change when viewed [with out whitespace differences](https://github.com/voodootikigod/node-serialport/compare/refactor?expand=1&w=1).


rwaldron:
LGTM, the tests pass, but we should probably do some reality checks first?

PR #718, opened 2016-04-03
Dependency cleanup (quick pass)

rwaldron:
- Eliminates use of node-pre-gyp as an actual module dependency (just a package dependency, this might even be just a devDep now?)
- Reduces 5 blocking fs operations to 1

Signed-off-by: Rick Waldron <waldron.rick@gmail.com>


rwaldron:
Thank you sir

PR #719, opened 2016-04-03
No need to use util.inherits unless actually using the super_ prop

rwaldron:
This will likely conflict with #718

Signed-off-by: Rick Waldron <waldron.rick@gmail.com>


rwaldron:
JFTR, this is exactly what the older Node.js runtimes are doing anyway, but I just looked and the newer runtimes are using `Object.setPrototypeOf(...)`, which is better than this approach, so I will close this

rwaldron:
For the kids at home: https://github.com/nodejs/node/commit/29da8cf8d7ab8f66b9091ab22664067d4468461e

PR #728, opened 2016-04-04
Fix double open and times when we don't close file descriptors for unix

reconbot:
- includes #670 from @gfcittolin
 - includes patch from @pabigot moving when we apply HUPCL
 - Close the fd if we have any errors during setup
 - Properly receive errors when setting a custom baud rate on OS X

### DIFF
--- a/src/serialport_unix.cpp
+++ b/src/serialport_unix.cpp
@@ -131,9 +131,6 @@ void EIO_Open(uv_work_t* req) {
   OpenBaton* data = static_cast<OpenBaton*>(req->data);
 
   int flags = (O_RDWR | O_NOCTTY | O_NONBLOCK | O_CLOEXEC | O_SYNC);
-  if(data->hupcl == false) {
-    flags &= ~HUPCL;
-  }
   int fd = open(data->path, flags);
 
   if(-1 == setup(fd, data)){
@@ -325,7 +322,9 @@ int setup(int fd, OpenBaton *data) {
 
   options.c_cflag |= CLOCAL; //ignore status lines
   options.c_cflag |= CREAD;  //enable receiver
-  options.c_cflag |= HUPCL;  //drop DTR (i.e. hangup) on close
+  if(data->hupcl) {
+    options.c_cflag |= HUPCL;  //drop DTR (i.e. hangup) on close
+  }
 
   // Raw output
   options.c_oflag = 0;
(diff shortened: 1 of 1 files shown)


pabigot:
I will test this later today or possibly tomorrow.

It is bad form to take material from pull requests, mix them together, and resubmit them under your own name, for two reasons:
* as @ecksun and I have both suggested, mixing multiple changes in one commit makes it hard to see what's going on and confirm that a problem has actually been fixed.
* Skilled users who take the time to identify, report, and provide tested fixes for problems see our solutions discarded or obscured.

This pull request could easily have been built by rebasing the patch from @gfcittolin and cherry-picking mine, then adding patches for the two additional issues you mention.  That approach would have avoided both those problems.

People like you and @rwaldron who provide open source benefit from free help.  People who provide contributions are compensated by seeing tools they care about improve and by receiving public acknowledgement through their commits being preserved in the project's history.  You do put at-mentions in the commit message, but that's a very limited form of traceability.

I can speak only for myself: my motivation to contribute to this project instead of simply forking it to meet my own needs is diminished by this sort of thing.

pabigot:
I do think splitting even this change set into smaller pieces is valuable as it makes the changes you've added obvious, and you've touched areas that still have problems.

In the original `EIO_Open` opens the file then passes it off to `setup`.  I believe a better approach is not to add `close()` in multiple places in `setup` but instead to detect that `setup` indicated an error and close the descriptor in the caller.  This puts the open and the close right next to each other.

The remaining question is what to do when `EIO_Update` invokes `setup` and discovers an error.  In the version you have in this PR the descriptor would be uniformly closed, which is a behavior change.  In the past somebody could recover by using a new configuration (at least if the rejection was from one of the code paths that didn't close the descriptor).  I suspect none of the problems discovered by setup() are fatal.  But if you prefer to have update close the underlying descriptor on an error, I think you'd have to also update `SerialPort.prototype.update` to record the closure in the object property.

Finally, I don't think your change in `EIO_Close` to remove the check for `EBADF` is correct, unless you've tracked all the code paths and are certain nobody would be able to reach that point when the device had already been closed.  However, the original code is certainly wrong.

I've pushed an alternative solution [here](https://github.com/pabigot/node-serialport/tree/serialport-unix-cleanup).  The commit messages in my additions explains more about why I made the changes, and I split them into two to make clear what changed versus what just moved.  I've only done minimal testing (confirmed no leaks), and I'm even more frightened by the code each time I look at it, so I'm not proposing that my version be merged without significant review and testing.  If anything in it or in what I've written here is helpful feel free to incorporate it, but I'm hoping to back away as this is starting to become a time sink for me.

PR #731, opened 2016-04-04
Update serialport_win.cpp (memory leak and infinite loop)

lobsangludd:
Fixed a memory leak in EIO_AfterWatchPort and infinite loop in EIO_Write in case of a disconnect (method never returns)


lobsangludd:
I see something similar was already committed 6 hours ago, sorry.

PR #740, opened 2016-04-08
Make the output of .list undefined if it’s an empty string

### LINKED ISSUE #735: Serialport.list has inconsistant representations for no data
Operating System and Hardware: all
NodeJS Version: all
Serialport version: 2.0.7-beta4

### Expected behavior
the same values for no data in the port description object

### Actual behavior
osx does empty string, linux does undefined, windows.. I don't know yet

### Steps and/or code to reproduce the issue
```js
require('serialport').list(console.log);
```

reconbot:
- Fixes #735 
 - Add some tests to start defining the bindings.

There has to be a better way to do the empty string checking.

### DIFF
--- a/src/serialport.cpp
+++ b/src/serialport.cpp
@@ -171,7 +171,7 @@ void EIO_AfterOpen(uv_work_t* req) {
 }
 
 NAN_METHOD(Update) {
-  
+
   // file descriptor
   if(!info[0]->IsInt32()) {
     Nan::ThrowTypeError("First argument must be an int");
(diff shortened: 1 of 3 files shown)


src/serialport.cpp
```diff
@@ -461,16 +461,52 @@ void EIO_AfterList(uv_work_t* req) {
...
+      if (strlen((*it)->productId.c_str()) > 0) {
+        Nan::Set(item, Nan::New<v8::String>("productId").ToLocalChecked(), Nan::New<v8::String>((*it)->productId.c_str()).ToLocalChecked());
+      } else {
+        Nan::Set(item, Nan::New<v8::String>("productId").ToLocalChecked(), Nan::Undefined());
+      }
+
       Nan::Set(results, i, item);
     }
-    argv[0] = Nan::Undefined();
+    argv[0] = Nan::Null();
```
rwaldron: Nice

rwaldron:
LGTM, but you may want someone with more NaN/V8 experience also sign off

PR #759, opened 2016-04-19
Stop bundling node-pre-gyp to allow for upgrades

### LINKED ISSUE #753: Remove node-pre-gyp as a bundled dependency and use preinstall script instead
According to the new docs [here](https://github.com/mapbox/node-pre-gyp#1-add-new-entries-to-your-packagejson), it might be considered better practice to add preinstall script `npm install node-pre-gyp` instead of using the `bundledDependencies` key.  This would allow other NPM modules to re-use the same version of node-pre-gyp, for example.

See the [node-pre-gyp Github repo](https://github.com/mapbox/node-pre-gyp) page for more info.  This issue might also be relevant: https://github.com/mapbox/node-pre-gyp/issues/162

reconbot:
- The latest and greatest recommendation from node-pre-gyp
 - Stops issues when people have different node-pre-gyp’s installed already

closes #753

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,10 +59,6 @@
     "optimist": "~0.6.1",
     "sf": "0.1.7"
   },
-  "bundledDependencies": [
-    "node-pre-gyp",
-    "node-pre-gyp-github"
-  ],
   "devDependencies": {
     "chai": "*",
     "chai-subset": "^1.2.2",
@@ -87,6 +83,7 @@
   },
   "license": "MIT",
   "scripts": {
+    "preinstall": "npm install node-pre-gyp node-pre-gyp-github",
     "install": "node-pre-gyp install --fallback-to-build",
     "rebuild": "npm rebuild && node-pre-gyp rebuild",
     "stress": "mocha --no-timeouts test/arduinoTest/stress.js",


package.json
```diff
@@ -87,6 +83,7 @@
   },
   "license": "MIT",
   "scripts": {
+    "preinstall": "npm install node-pre-gyp node-pre-gyp-github",
     "install": "node-pre-gyp install --fallback-to-build",
     "rebuild": "npm rebuild && node-pre-gyp rebuild",
```
springmeyer: I'm curious: why are you overriding `rebuild` that way?

package.json
```diff
@@ -87,6 +83,7 @@
   },
   "license": "MIT",
   "scripts": {
+    "preinstall": "npm install node-pre-gyp node-pre-gyp-github",
```
springmeyer: It should be harmless but I doubt you need to have `node-pre-gyp-github` in the `preinstall` deps. I presume it can just be a normal dep since it only needs to be used by developers when posting new binaries (not when installing), if I'm understanding its purpose correctly.

reconbot: It's needed to download the binaries

package.json
reconbot: this was unintentional

package.json
springmeyer: Are you sure? From the readme (https://github.com/bchr02/node-pre-gyp-github) it looks like it is only for posting binaries. As `node-pre-gyp` itself downloads over http, it handles that fine.

PR #763, opened 2016-04-25
Refactor Unix setting BaudRates

reconbot:
I moved the baud rate stuff to another function, even though we'll end up applying settings to a port more often then strictly necessary I think this is a step in the right direction. The cost there is really small.

I'm really new at C++ so I'm not sure if I'm causing memory leaks and the like.


src/serialport_win.cpp
```diff
@@ -170,7 +170,9 @@ struct WatchPortBaton {
 };
 
 void EIO_Update(uv_work_t* req) {
-  // TODO always error or do an update
+  ConnectionOptionsBaton* data = static_cast<ConnectionOptionsBaton*>(req->data);
+  _snprintf(data->errorString, sizeof(data->errorString), "Update is not yet implemented on windows");
+  // https://msdn.microsoft.com/en-us/library/windows/desktop/aa363260(v=vs.85).aspx
```
reconbot: I'd love to land this patch with windows support.

src/serialport_unix.cpp
```diff
@@ -141,103 +141,111 @@ void EIO_Open(uv_work_t* req) {
...
-  if (-1 == setup(fd, data)) {
-    return;
+  // get port options
+  struct termios options;
+  tcgetattr(fd, &options);
+
+  // If there is a custom baud rate on linux you can do the following trick with B38400
+  // Might be an outdated way of setting custom rates https://sourceforge.net/p/pyserial/patches/28/
+  // termios2 is now a thing
+  #if defined(__linux__) && defined(ASYNC_SPD_CUST)
```
reconbot: Looks like if we use the new struct we'll be able to land support for any supported rate for linux

src/serialport_win.cpp
```diff
@@ -64,7 +65,7 @@ void EIO_Open(uv_work_t* req) {
   if (file == INVALID_HANDLE_VALUE) {
     DWORD errorCode = GetLastError();
     char temp[100];
-    _snprintf(temp, sizeof(temp), "Opening %s", data->path);
+    _snprintf(temp, sizeof(temp), "Opening %s", origionalPath);
```
reconbot: This might be a separate PR

src/serialport.h
```diff
@@ -64,11 +64,9 @@ struct OpenBatonPlatformOptions { };
 OpenBatonPlatformOptions* ParsePlatformOptions(const v8::Local<v8::Object>& options);
 
 struct OpenBaton {
-  char path[1024];
```
reconbot: only sorting these so it's clear what needs to be deleted after use

src/serialport_unix.cpp
```diff
@@ -372,7 +366,7 @@ void EIO_Write(uv_work_t* req) {
 void EIO_Close(uv_work_t* req) {
   CloseBaton* data = static_cast<CloseBaton*>(req->data);
   if (-1 == close(data->fd)) {
-    snprintf(data->errorString, sizeof(data->errorString), "Unable to close fd %d, errno: %d", data->fd, errno);
+    snprintf(data->errorString, sizeof(data->errorString), "Error: %s unable to close fd %d", strerror(errno), data->fd);
```
reconbot: this might be a separate pr

test/arduinoTest/arduinoEcho/arduinoEcho.ino
```diff
@@ -1,13 +1,43 @@
+// Echo all ASCII characters
+// Allow switching of baud rates
+
```
reconbot: I want this to be part of an integration test for testing baud rate changes

Issue #702, opened 2016-03-28
Disconnect is poorly defined and doesn't always work


reconbot:
We call `disconnected` from all over the place, c++ bindings call it from the options. Occasionally we remove all events. Sometimes we emit the close event. It's also not documented.

Having this be the discussion issue around how it currently works and how we should fix it up.

fivdi:
@reconbot should the issue title be "Disconnect is poorly defined and *doesn't* always work"?

reconbot:
yep =)

reconbot:
We now have a single disconnect path and never remove events. I think ... I think this is closable. 

While there are internal cleanups to be had, the api is now documented and consistent.

PR #764, opened 2016-04-25
Stop removing all event listeners on close and disconnect

### LINKED ISSUE #638: Calling `close()` removes all event listeners (including "error")
When a SerialPort is closed by calling `close()`, all event listeners are removed (as per the documentation).  This behavior might be undesired.

If one calls `write` on a closed serial port connection without passing a callback to the `write` method, an "error" event is emitted by default.  So, calling `port.write(buf);` on a closed serial port can cause an "error" event to be emitted without an "error" event handler (even if the user created an "error" event handler before the SerialPort was closed).  In this case, the default Node behavior for this is to throw an exception, which could crash the program.

reconbot:
Should address some of #702 and closes #638 and helps https://github.com/rwaldron/johnny-five/issues/1037

### DIFF
--- a/lib/serialport.js
+++ b/lib/serialport.js
@@ -410,9 +410,6 @@ SerialPort.prototype.disconnected = function (err) {
     debug('Disconnect completed with an exception: ' + JSON.stringify(e));
   }
 
-  // TODO THIS IS CRAZY TOWN
-  self.removeAllListeners();
-
   self.closing = false;
   self.fd = null;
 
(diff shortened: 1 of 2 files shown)


lib/serialport.js
```diff
@@ -410,9 +410,6 @@ SerialPort.prototype.disconnected = function (err) {
     debug('Disconnect completed with an exception: ' + JSON.stringify(e));
   }
 
-  // TODO THIS IS CRAZY TOWN
```
bcjordan: 😆

PR #767, opened 2016-04-26
Drop some dependencies for a smaller bundle size

reconbot:
- Drop `bluebird` 168kb for `es6-promise` 37.1kb (vs 18.1kb library code) bundled sizes
 - Move `node-pre-gyp-github` to dev deps so it's not always installed (5.8MB package size)
 - Stop bundling `node-pre-gyp-github` as it's only needed during publishing


rwaldron:
LGTM

Issue #747, opened 2016-04-08
SerialPort in Galileo Gen 2 running Yocto


ortizvinicius:
I'm trying to set a Serial Communication in a Intel Galileo Gen 2 with Linux Yocto, but I only get the message "**Illegal Instruction**".

Issue opened before in the wrong place: https://github.com/rwaldron/johnny-five/issues/1086

The code:

```
var Serialport = require("serialport").SerialPort;
var sp = new Serialport("/dev/ttyS0", {
  baudRate: 57600
});

sp.on("open", function() {
  console.log("Port is open!");

  // Once the port is open, you may read or write to it.
  sp.on("data", function(data) {
    console.log("Received: ", data);
  });

  sp.write(new Buffer("0120000003"));
});
```

Already tried to use the last beta version of SerialPort.

ortizvinicius:
To make sure i'm not doing anything wrong:
- i got a new sd card with yocto image
- updated the node with this: (don't know if this could make a difference):
```
npm cache clean -f
npm install -g n
n stable
```
- instaled the serialport module
- and done this in terminal:
```
cd /sys/class/gpio                                                                         
echo -n "28" > export
echo -n "32" > export
echo -n "45" > export
cd gpio28
echo -n "out" > direction
echo -n "0" > value
cd gpio32
echo -n "out" > direction
echo -n "1" > value
cd gpio45
echo -n "out" > direction
echo -n "1" > value
stty -F /dev/ttyS0 57600
```

ortizvinicius:
I think node-pre-gyp rebuild worked, but i can only test tomorrow, and keep you updated

fivdi:
Yes, `node-pre-gyp rebuild` works around the `Illegal Instruction` issue.

It's definitely the case that pre-compiled binaries that result in `Illegal Instruction` errors are being installed initially:
```
root@galileo:~/serialport# npm install serialport
|
> serialport@2.0.7-beta5 install /home/root/serialport/node_modules/serialport
> node-pre-gyp install --fallback-to-build

[serialport] Success: "/home/root/serialport/node_modules/serialport/build/Release/serialport.node" is installed via remote
serialport@2.0.7-beta5 node_modules/serialport
├── bindings@1.2.1
├── sf@0.1.7
├── async@0.9.0
├── debug@2.2.0 (ms@0.7.1)
├── nan@2.2.1
[...]

PR #779, opened 2016-05-02
Disable instruction sets when building for linux on ia32 for Galileo support

reconbot:
Per #747 some of the intel boards don't support some instructions and our prebuilt binaries wont run on those platforms. This PR disables a few instruction sets for 32bit linux installations. If these instructions are available but disabled it's a performance hit, if they're not available but enabled they cause an annoying error with little information. I don't have data for what kind of performance hit. I'm actually just assuming.

We can do one of a few things.
- Not provide prebuilt binaries for linux/ia32 (It's pretty heavily downloaded see the issue for counts)
- Do what this PR does and disable the instructions for everybody
- Something better

cc @rwaldron @jacobrosenthal @fivdi

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -3,7 +3,7 @@
     {
       'target_name': 'serialport',
       'sources': [
-        'src/serialport.cpp',
+        'src/serialport.cpp'
       ],
       'include_dirs': [
         '<!(node -e "require(\'nan\')")'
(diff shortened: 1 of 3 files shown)


fivdi:
A potential alternative not listed in the OP is to do nothing. It looks like there was only one issue about the topic filed at the serialport repo so far. The workaround on the Galileo 2 would be `npm install serialport --build-from-source`.

PR #782, opened 2016-05-03
Build precompiled binary for Node.js 4.4 on Windows

deadprogram:
Build pre-compiled binary for Node.js 4.4 on Windows instead of 4.3, now that 4.4.3 is the LTS version.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,7 +15,7 @@ environment:
   matrix:
   - nodejs_version: "0.10"
   - nodejs_version: "0.12"
-  - nodejs_version: "4.3"
+  - nodejs_version: "4.4"
   - nodejs_version: "5.11"
   - nodejs_version: "6.0"
 


deadprogram:
Hi @reconbot seems like the problem is related to npm as discussed here: http://help.appveyor.com/discussions/problems/3147-error-eperm-operation-not-permitted-rename-on-npm-install

Updating the Windows build to install a newer npm seems like it could be a relatively easy fix. What do you think?

deadprogram:
I think you need to do this: https://github.com/npm/npm/wiki/Troubleshooting#upgrading-on-windows

deadprogram:
Looks like all roads lead back to node-gyp https://github.com/mapbox/node-pre-gyp/issues/209

deadprogram:
Thanks @reconbot I that case I'm deleting this branch due to subsequent branch.

PR #806, opened 2016-05-14
Refactor list-unix.js in order to use libudev directly

Zensey:
This patch simplifies listUnix by means of libudev binding.
It reduces number of dependencies and cpu consumption.
Now listUnix needs no to checkPathAndDevice each device. With libudev you simpy get all devies from subsystem that you need (for ex. "tty").


Zensey:
> I also know we currently show more than the tty subsystem.

If you mean bluetooth serial ports -- rfcomm* -- they are from "tty" subsystem, too.

Zensey:
So we need to provide precompiled binaries of udev.
Are there documents / tutorials about this ?

Zensey:
If you mean the requirement to provide precompiled binaries for various platforms, then it seems reasonable to include code in place of stub. 
I had the same idea.

Zensey:
ok

PR #808, opened 2016-05-15
write binding tests and cleanup

reconbot:
- Fix bug where write callback was being called multiple times when write operations blocked
 - [windows] refactored write code to be less complex
 - added arduino required integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,7 +364,7 @@ The `buffer` parameter accepts a [`Buffer` ](http://nodejs.org/api/buffer.html) 
 
 **_callback (optional)_**
 
-Called once the write operation returns. The callback should be a function that looks like: `function (error, bytesWritten) { ... }`
+Called once the write operation returns. The callback should be a function that looks like: `function (error) { ... }`
 
 **Note:** The write operation is non-blocking. When it returns, data may still have not actually been written to the serial port. See `drain()`.
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "serialport",
-  "version": "3.1.2",
+  "version": "3.2.0-beta1",
   "description": "Node.js package to access serial ports. Welcome your robotic javascript overlords. Better yet, program them!",
   "author": {
     "name": "Chris Williams",
@@ -10,7 +10,7 @@
   "binary": {
     "module_name": "serialport",
     "module_path": "build/{configuration}/",
-    "host": "https://github.com/voodootikigod/node-serialport/releases/download/3.1.2"
+    "host": "https://github.com/voodootikigod/node-serialport/releases/download/3.2.0-beta1"
   },
   "main": "./lib/serialport",
   "repository": {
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -364,7 +364,7 @@ The `buffer` parameter accepts a [`Buffer` ](http://nodejs.org/api/buffer.html)
 
 **_callback (optional)_**
 
-Called once the write operation returns. The callback should be a function that looks like: `function (error, bytesWritten) { ... }`
+Called once the write operation returns. The callback should be a function that looks like: `function (error) { ... }`
```
reconbot: I've removed the bytes written parameter from the callback as it was usually a lie because the callback was called over and over. It should always match the length of the bytes you gave it to write. It's possible to put this back but not easy and not useful?

package.json
```diff
@@ -92,6 +92,7 @@
     "rebuild-all": "npm rebuild && node-gyp rebuild",
     "gyp-rebuild": "node-gyp rebuild",
     "stress": "mocha --no-timeouts test/arduinoTest/stress.js",
+    "integration": "mocha test/arduinoTest/integration.js",
```
reconbot: This is a big change, we now have a small suite of tests that require an arduino

src/serialport.cpp
```diff
@@ -321,6 +319,8 @@ void EIO_AfterWrite(uv_work_t* req) {
   // remove this one from the list
   queuedWrite->remove();
 
+  data->callback->Call(1, argv);
```
reconbot: Now only call the callback when we're done with the whole write operation and have checked for errors that we should never hit. We shouldn't be throwing on th previous lines but if we ever do it might be worth ending the program because something is way out of wack. I'm not even sure we need the check.

src/serialport_win.cpp
```diff
@@ -57,11 +57,11 @@ void EIO_Open(uv_work_t* req) {
   HANDLE file = CreateFile(
     data->path,
     GENERIC_READ | GENERIC_WRITE,
-    0,
+    0, // dwShareMode 0 Prevents other processes from opening if they request delete, read, or write access
     NULL,
     OPEN_EXISTING,
-    FILE_FLAG_OVERLAPPED,
-    NULL);
```
reconbot: removed optional argument and commented what the other args mean

src/serialport_win.cpp
```diff
@@ -368,35 +368,31 @@ void EIO_Write(uv_work_t* req) {
...
-      // Write operation completed synchronously
-      data->result = bytesWrittenSync;
-    }
+      // Write operation is completing asynchronously
+      // We MUST wait for the operation completion before deallocation of OVERLAPPED struct
+      // or write data buffer
 
+      // block for async write operation completion
+      bytesWritten = 0;
+      if (!GetOverlappedResult((HANDLE)data->fd, &ov, &bytesWritten, TRUE)) {
```
reconbot: This got me at first. Why are we opening the port for async if we always block when writing? Because if we don't it would serialize reading and writing, and they'd block each other. That's why.

PR #822, opened 2016-05-24
Remove the serial port factory and throw errors on invalid arguments

### LINKED ISSUE #820: Throw argument errors during construction instead of passing them in a callback
Right now options errors when creating a serialport happen in a callback that is called immediately if provided to the constructor or emitted on the factory object. Lets throw these errors immediately. 

We won't need the factory after this which things even further.

I suggest we return the constructor, but keep a copy of it on a `SerialPort` property as to not confuse too many people with this change. (eg, most example code on the net will continue to work) which we can depreciate later. `list` and `parsers` will also need to be properties of the constructor function too.

reconbot:
- Constructor now throws errors when given invalid options
 - The Constructor’s callback is given to `.open()` if `openImmediately` is set
 - Depreciation warning when using the old style SerialPort.SerialPort

Closes #820

### DIFF
--- a/bin/serialport-terminal.js
+++ b/bin/serialport-terminal.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 'use strict';
 
-var serialport = require('../');
+var SerialPort = require('../');
 var version = require('../package.json').version;
 var args = require('commander');
 
@@ -21,7 +21,7 @@ args
   .parse(process.argv);
 
 function listPorts() {
-  serialport.list(function(err, ports) {
+  SerialPort.list(function(err, ports) {
     if (err) {
       console.error('Error listing ports', err);
     } else {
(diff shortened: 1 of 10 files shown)


bin/serialport-terminal.js
```diff
@@ -49,7 +49,7 @@ var openOptions = {
   stopBits: args.stopbits
 };
 
-var port = new serialport.SerialPort(args.port, openOptions);
+var port = new Serialport(args.port, openOptions);
```
gnarf: Just to mention, you might want to hunt for `Serialport` -> `SerialPort` for consistency

lib/serialport.js
```diff
@@ -493,4 +478,10 @@ SerialPort.prototype.drain = function(callback) {
   }.bind(this));
 };
 
-module.exports = factory;
+SerialPort.parsers = parsers;
+SerialPort.list = SerialPortBinding.list;
+
+// TODO deprecate
+SerialPort.SerialPort = SerialPort;
```
gnarf: Something like? 
```js
Object.defineProperty(SerialPort, 'SerialPort', {
  get: function() { 
    console.warn('DEPRECATION: Please use `require(\'serialport\')` instead of `require(\'serialport\').SerialPort`');
    Object.defineProperty(SerialPort, 'SerialPort', { value: SerialPort });
    return SerialPort;
  },
  configurable: true
});
```

PR #830, opened 2016-05-28
Remove the openImmediatly flag in favor of the autoOpen option

### LINKED ISSUE #809: Move the `openImmediately` flag to the port options.
Having this boolean in the constructor isn't nearly as clean as having it in the options. Lets move it to the options and keep it's default behavior to `true`. This would be a major api change so we should bundle it with a few other api changes.

reconbot:
closes #809

`autoOpen` makes a lot more sense to me. What do you think?

### DIFF
--- a/lib/serialport.js
+++ b/lib/serialport.js
@@ -34,6 +34,7 @@ var kMinPoolSpace = 128;
 
 var defaultSettings = {
   baudRate: 9600,
+  autoOpen: true,
   parity: 'none',
   xon: false,
   xoff: false,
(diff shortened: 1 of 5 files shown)


lib/serialport.js
```diff
@@ -76,19 +77,18 @@ function correctOptions(options) {
...
 }
 
-function SerialPort(path, options, openImmediately, callback) {
-  var args = Array.prototype.slice.call(arguments);
-  callback = args.pop();
-  if (typeof callback !== 'function') {
-    callback = null;
+function SerialPort(path, options, callback) {
+  if (typeof callback === 'boolean') {
+    throw new TypeError('`openImmediately` is now called `autoOpen` and is a property of options');
```
reconbot: I can see people not reading the upgrade notes for a while and this should help prevent issues with that.

PR #890, opened 2016-07-28
Byte delimiter should emit only when delimiters strictly follow each other

arve0:
Example:

```
var parser = byteDelimiter([0,0]);

parser(emit, new Buffer('\0hello\0world\0\0');
```

Expected: `emit` should be called once, upon `\0\0` with whole string.
What happens: `emit` is called twice, with `\0hello\0` and `world\0\0`.

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -2,6 +2,20 @@
 
 // Copyright 2011 Chris Williams <chris@iterativedesigns.com>
 
+function isEqual(arr1, arr2) {
+  var isArray = Array.isArray;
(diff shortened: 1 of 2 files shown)


arve0:
Hm, this pattern will not work out:

```
var parser = byteDelimiter([0,1]);

parser(emit, new Buffer('hello\0\0\1');
```

arve0:
> I'll need to consider if this bug was a "feature"

Yes, the same stroke me.

> if this should be considered a breaking change. 

Yes, I would consider this a breaking change.

> Seems like a small issue for a 5.x release.

Paper is free when drawing on iPad, it's just bits and bytes 🙈 No rush here, though.

PR #897, opened 2016-08-06
Start to build an auto flashing Arduino pipeline

reconbot:
This will give the commands necessary to flash a connected arduino with the latest test firmware.


package.json
```diff
@@ -87,8 +89,11 @@
   },
   "license": "MIT",
   "scripts": {
+    "arduino:compile": "avr-pizza compile -a uno -s test/arduinoTest/arduinoEcho/arduinoEcho.ino -o test/arduinoTest/arduinoEcho",
+    "arduino:diff": "echo 'help me noopkat! && false",
```
noopkat: bait taken 😁  what is this part and can I help?

reconbot: Oh man, I probably should have committed this. Alright, the idea here was to ensure people committed the compiled version of the script and have a test that would fail if you didn't.

I'm not sure this was the right idea. What do you think?

noopkat: Oh! That makes sense, no worries. This would involve running avr-pizza programatically, then using `fs` to read the existing `test/arduinoTest/arduinoEcho`, and then comparing contents. 

Alternatively, this might be a good npm pre-publish hook, or git pre-commit hook to just use avr-pizza to produce the file automatically for the person.

noopkat: this is rather terribly written, but [here's a gist](https://gist.github.com/noopkat/f380a22671d2fb18f16157514219358a) as a quick example.

PR #906, opened 2016-08-14
Streams rewrite

reconbot:
This is a major change.  To reading, the poller and the bindings layer, there isn’t a great way to do this piecemeal. The goal here is to make bindings a plugin layer, our c++ bindings isolated and tested, and make Serialport a streams object.

- Move all bindings into platform specific modules DarwinBinding LinuxBinding, and WindowsBinding
- Separate out the bindings from the SerialPort class so it can be required by itself.
- Provide mock bindings as a tested first class bindings layer
- Mixin stream methods with `pushBindingWrap`, inspired from fs.readStream
- All integration tests now require an Arduino 
- Ditch sandboxed modules as we don't need to be that heavy handed to test anymore

TODO

- windows binding
- performance test
- memory leak test
- 🎉


README.md
```diff
@@ -52,9 +52,7 @@ For getting started with node-serialport, we recommend you begin with the follow
     * _instance_
         * [`.open([callback])`](#module_serialport--SerialPort+open)
         * [`.update([options], [callback])`](#module_serialport--SerialPort+update)
-        * [`.write(data, [callback])`](#module_serialport--SerialPort+write)
-        * [`.pause()`](#module_serialport--SerialPort+pause)
-        * [`.resume()`](#module_serialport--SerialPort+resume)
+        * [`._write(data, [callback])`](#module_serialport--SerialPort+_write)
```
mcollina: Given this is now a stream, `_write` should probably not be documented.

munyirik:
Thanks @reconbot. I'll take a look.

munyirik:
Looks like it would fit nicely. When will you be merging this?

vicatcu:
@reconbot lol, congrats dude!

munyirik:
@reconbot congrats!!

mcollina:
Woow, good work! There seems to be a couple of spurious issues on Travis (sigh).

I'm eager to see this out!

jupe:
should this affect also https://github.com/EmergingTechnologyAdvisors/node-serialport/pull/923 memory leak issue ?

Issue #545, opened 2015-06-25
250k baudrate


Heash:
Now supported baudrates are: 115200, 57600, 38400, 19200, 9600, 4800, 2400, 1800, 1200, 600, 300, 200, 150, 134, 110, 75, or 50

Pyerial lib has patch for supporting 250000 baudrate. But it will be very nice if node-serialport will suport this too.

jacobrosenthal:
PR Welcome. Wed need to investigate if that baud exists on osx and win as well.

ChrisTerBeke:
What is the status of this? We're really waiting for 250k baudrate support on linux. More and more devices use this baudrate to have faster communication nowadays (including a lot of IoT stuff).

jacobrosenthal:
PRs still welcome.

rwaldron:
@jacobrosenthal I was checking out the PySerial code and they just raise a ValueError for platforms that don't support. The patch for node-serialport should just throw a RangeError

ChrisTerBeke:
We've put some time in it to see what's going on and it seems very possible to do (latest linux kernels already support it), but sometimes the linux header files with correct baudrate info are missing so compiling throws an error on that. Also, the custom baudrate option in node-serialport seems interesting. Maybe we can adapt that functionality a bit to allow for more dynamic baudrates without having to compile for it (ofc we need to make sure that the clock dividers are set correctly)?

ChrisTerBeke:
We gave up on this and wrote our own drivers in C/C++ and made a node addon out of it.

farfromrefug:
@ChrisTerBeke you wrote an alternative to ```serialport```? is it open source?

ChrisTerBeke:
@farfromrefug it's kinda part of a larger application that is not open source, but if there's enough demand, we might look into open sourcing the serial comm part. It works on Linux, OSX and Windows (only confirmed on win10). Anyone interested?

reconbot:
We're going to land 250000 pretty soon. If you have any code that would help cleanup or improve node serialport, that would be wonderful.

reconbot:
Specifically, I've started refactoring the baudrate code, but have yet to land windows support for updating the baudrate. I'd love help on this PR https://github.com/voodootikigod/node-serialport/pull/763

farfromrefug:
@ChrisTerBeke If that s ok with you i think it would be awesome for you to share that project. Not to create a side project from ```serialport``` but to actually help improving it.
[...]

PR #918, opened 2016-08-29
Issue #545: 250k baudrate 

fnipo:
It is a workaround that I have made to make custom baudrates work in a debian linux.
These changes include changes in the _serialport_unix.cpp_ file to use the new unix serial interface, through the _termios2_ structure. Moreover, problems may occur during build due to the fact that some _libuv_ files are linked to the old _termios_ structure. To solve it, I needed to do a one line change in the _uv-unix.h_ file of the _.node-gyp_ folder to make the build works. To simplify the build process, I create the _BUILD.sh_ script that change the _libuv_ file automatically before build, and reverse the change after build. So this module is supposed to build only using: **_sudo ./BUILD.sh_**.
I have tested It, and it is working for 250k baudrate communications with a Ultimaker 3D printer.


fnipo:
Hi @reconbot,

Sorry for the readme file update, I rollbacked it.
Regarding the node patching, I just need to do a modification to support other node versions.
Is it enough? Do you suggest any other modification in the build script?

PR #922, opened 2016-09-04
Transform parsers into streams

reconbot:
Parsers are now transform streams!

I'm not 100% convinced this is easier to use but it is more clear about the mechanics. I think that might make it worthwhile. It also opens up a clear example to how to make your own parser.


README.md
```diff
@@ -601,6 +543,65 @@ SerialPort.list(function (err, ports) {
...
+```js
+var SerialPort = require('serialport');
+var port = new SerialPort('/dev/tty-usbserial1');
+var parser = new SerialPort.parsers.ReadLine();
+port.pipe(parser);
+parser.on('data', console.log);
+port.write('ROBOT PLEASE RESPOND\n');
+
+// this can be shortened to
+var SerialPort = require('serialport');
```
reconbot: This is odd/wrong

PR #923, opened 2016-09-07
[Question] Trying to tackle the memory leak..

RomanSaveljev:
Hello,

We are experiencing RSS going out of control in our application using `serialport` module. After battling the issue for a long we have come to the conclusion it is due to native part of the module.

The enclosed simple application `t.js` shows problem manifesting itself.

[There](https://asciinema.org/a/ac16a5x8van1ftwu8rl1woj9l) is some test captured with asciinema: upper pane runs `t.js` against unpatched code and the below is checking this patch.

Perhaps, our test application is using the library in a wrong way. It would be nice to know how to fix it.

### DIFF
--- a/t.js
+++ b/t.js
@@ -1,0 +1,69 @@
+var _ = require('lodash');
+var Promise = require('bluebird');
+var serialport = require('./lib/serialport');
+var heapdump = require('heapdump');
+
+var port = process.env.TEST_PORT;
+
+if (!port) {
+  console.error('Please pass TEST_PORT environment variable');
+  process.exit(1);
+}
+
+setInterval(function() {
+	typeof gc !== 'undefined' && console.log('forcing GC') && gc();
+	console.log(process.memoryUsage());
+}, 1000);
+
+function dumpHeap() {
+	console.log('Dumping heap');
+	heapdump.writeSnapshot();
+}
+
+setTimeout(dumpHeap, 6000);
+setInterval(dumpHeap, 30000);
+
+var counter = 0;
+
+function makePromise() {
+	var self = {};
+	return new Promise(function(resolve, reject) {
+		if (counter++ % 100 == 0) {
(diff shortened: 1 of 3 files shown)


RomanSaveljev:
Thanks for getting back to us. We are too familiar how these memory leaks can be tricky to pinpoint, so have tried to be as precise as possible.

Here is the test system details:

```
e107752:tmp romsav01$ uname -a
Darwin romsav01-mac.lan 15.6.0 Darwin Kernel Version 15.6.0: Mon Aug 29 20:21:34 PDT 2016; root:xnu-3248.60.11~1/RELEASE_X86_64 x86_64
e107752:tmp romsav01$ node --version
v6.2.1
```

By the way, it was possible to unlock the `SerialPoller` into being collectable by putting `delete serport.serialPoller` to "close" event handler. But then in the production test (conditions are unknown) we began to see node crashes like this:

```
Error in `node': free(): invalid pointer: 0x0172ffec 
```

Seems like some memory is being freed twice now

t.js
```diff
@@ -0,0 +1,69 @@
...
+		};
+		self.serialPort = new serialport(port, options);
+			function Oo(c) {
+			this.counter = c;
+			this.arr = "thisisstring";
+			for (var i = 0; i < 100; i++) {
+				this.arr += i + c;
+			}
+		}
+		self.serialPort.on('open', resolve.bind(new Oo(counter)));
```
jupe: `Oo` ?

jupe:
probably that `Oo` is not important. Are you @reconbot able to reproduce issue?

RomanSaveljev:
`Oo` just adds to bound context to see memory consumption grow faster. It is not important for the testcase itself

N4TY:
unsubscribe

jupe:
any estimates when we could get fix for this? seems that this is quite fatal problem at the moment for us..

PR #936, opened 2016-09-19
Drop explicit support for Node.js v5

reconbot:
Node.js v5 has been dropped by node core as it is not an LTS version.

We still support 0.10 and 0.12 because there are a few devices out there who absolutely cannot upgrade. (The begalbone black is one of them.) When this changes or if their use drops we'll drop support for pre Node 4.0.

### DIFF
--- a/.docs/README.hbs
+++ b/.docs/README.hbs
@@ -60,18 +60,18 @@ For getting started with node-serialport, we recommend you begin with the follow
 ## Platform Support
 `serialport` supports and tests against the following platforms, architectures and node versions.
 
-| Platform / Arch | Node v0.10.x | Node v0.12.x | Node v4.x | Node v5.x | Node v6.x |
-|       ---       | --- | --- | --- | --- | --- |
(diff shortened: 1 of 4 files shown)


rwaldron:
> (The begalbone black is one of them.) 

I recently learned that [node-i2c](https://github.com/kelly/node-i2c) is preventing upgrade. Here's the dep flow: 

1. BoneScript depends on node-i2c
2. node-i2c appears to be unmaintained, doesn't support Node > 0.12.0 (regardless of what the engines property says)


Even more unfortunate...


1. Beaglebone-IO depends on BoneScript
2. Johnny-Five must respect the minimum viable node version of all its IO Plugins
3. Beaglebone-IO is preventing Johnny-Five from dropping Node 0.12.0, because BoneScript is preventing Beaglebone-IO from dropping Node 0.12.0, because node-i2c is preventing BoneScript from dropping Node 0.12.0, because the maintainer is not maintaining. 

How about that for bullshit?

PR #940, opened 2016-09-24
UWP support

munyirik:
This change adds support for serialport to work in [UWP](https://en.wikipedia.org/wiki/Universal_Windows_Platform) apps (aka Windows Store apps).

It doesn't affect existing code since [special flags](https://github.com/ms-iot/ntvsiot#installing-npm-packages) need to be passed to npm for serialport_uwp.cpp to build (the version of npm that supports the UWP flags is in the [chakra branch](https://github.com/ms-iot/node)).

I'm still working on getting tests set up for CI but so far you can see the results from running bindings.js and integration-lite.js below:
[sp-uwp-test-results.txt](https://github.com/EmergingTechnologyAdvisors/node-serialport/files/490858/sp-uwp-test-results.txt)

The goal for this PR is to remove the step required to 'patch' serialport after installing it when targeting UWP. For example [see the steps](https://developer.microsoft.com/en-us/windows/iot/samples/j5servocontroller) to build a Johnny-Five app. After 'npm install serialport', serialport.node needs to be replaced with the one built for UWP.

The API's used in the new file are from https://msdn.microsoft.com/library/windows/apps/windows.devices.serialcommunication.serialdevice.aspx


src/serialport.h
```diff
@@ -150,6 +150,9 @@ struct ListResultItem {
   std::string locationId;
   std::string vendorId;
   std::string productId;
+#ifdef UWP_DLL
+  std::string displayName;
```
munyirik: I'll take it out so it's consistent with other platforms. 
The displayName is supposed to be the friendly name for the device. It maps to the [System.ItemNameDisplay](https://msdn.microsoft.com/en-us/library/windows/desktop/bb760770(v=vs.85).aspx) Windows property. I thought it would be nice to have but it's not necessary.
So for example if I have a Leonardo attached to my machine and I run serialport.list, this is what I can get...

>comName: COM4
pnpId: USB\VID_2341&PID_8036&MI_00\7&1f38248&0&0000
manufacturer: undefined
displayName: Arduino Leonardo

reconbot: It's not a bad thing to have, if it's not a Windows only thing I'm for
adding it but I don't know where it comes from.

munyirik:
Will do. I haven't had a chance to get back to this but I will in the next couple of days. I'll let you know if I have questions. Thanks!

munyirik:
Update: I'd like to get the tests running on a real machine but I can't do that currently with appveyor. They do say that they are [already working on that feature](http://help.appveyor.com/discussions/questions/2779-using-external-machine-to-buildtest) but no ETA.

munyirik:
Hey @reconbot. It's on the radar. There's a bunch of work that's fallen behind related to the UWP project so that needs to be addressed first. No ETA unfortunately.

cc. @paulmon, @msalehmsft

PR #952, opened 2016-10-05
Prevent exception in Buffer.concat, when data starts coming too fast

jupe:
Fix issue like this:
```
TypeError: Cannot read property 'length' of undefined
    at Function.Buffer.concat (buffer.js:306:24)
    at SerialPort.<anonymous> (/../node_modules/serialport/lib/serialport.js:348:32)
    at SerialPort.<anonymous> (/../node_modules/serialport/lib/serialport.js:364:7)
    at FSReqWrap.wrapper [as oncomplete] (fs.js:614:17)
uncaughtException: Cannot read property 'length' of undefined
```

### DIFF
--- a/lib/serialport.js
+++ b/lib/serialport.js
@@ -284,7 +284,11 @@ if (process.platform !== 'win32') {
 
         // do not emit events if the stream is paused
         if (this.paused) {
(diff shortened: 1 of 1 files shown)


jupe:
seems that branch was already obsolete, but fix is still valid..

PR #976, opened 2016-10-21
put the dataCallback back for windows reads

### LINKED ISSUE #969: v5.0.0-beta1 broken on Windows; `TypeError: undefined is not a function`
SerialPort version: 5.0.0-beta1
NodeJS Version: 6.9.0
Operating System and Hardware: Windows 10 Pro 64-bit, Lenovo W520 laptop
Are you using Electron/Atom/NwJS?: No (not to demonstrate the bug)

```
$ node -v && npm -v && python --version
v6.9.0
3.8.6
Python 2.7.11
```

### Summary of Problem
In windows, when data is received on a serial port the `data` callback never fires, and even if there is no handler assigned for the `data` event the following appears in the output regardless:
```
undefined:0



TypeError: undefined is not a function
```

### Expected Behavior
No error should be caused just by receiving data and all `data` callbacks should fire. This seems to work in version 4.0.3. 

### Actual Behavior
Error occurs and callback does not fire, as described in summary above.

Open a port and send data to it (i.e. externally).
Simple demo code available here:
https://github.com/jacobq/serialport-5.0.0-beta1-bug-demo

reconbot:
closes #969

### DIFF
--- a/lib/serialport.js
+++ b/lib/serialport.js
@@ -164,6 +164,8 @@ function SerialPort(path, options, callback) {
     this.bufferSize = settings.bufferSize;
     this.readable = true;
     this.reading = false;
+  } else {
+    settings.dataCallback = this._emitData.bind(this);
   }
 
   this.options = settings;
@@ -293,6 +295,11 @@ SerialPort.prototype.write = function(data, callback) {
   }.bind(this));
 };
 
+SerialPort.prototype._emitData = function(data) {
+  this.emit('data', data);
+};
+
+
 if (process.platform !== 'win32') {
   SerialPort.prototype._read = function() {
     if (!this.readable || this.paused || this.reading || this.closing) {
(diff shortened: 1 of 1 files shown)


jacobq:
Is there a way to make a regression test for this? (e.g. make serial port think it received data) If someone would help me I'd be willing to attempt to dive in an learn. We are now using this library in commercial, production desktop app, so I am rather committed :)

jacobq:
No, but I'll put it at the top of my things to work on when I need a break from my usual routine.

jacobq:
So I haven't worked on getting CI code to test yet, but I did find that the current master (and presumably all prior 5.x versions) does not work in electron on Windows. It works OK from REPL, but in electron if you open the port and write it sends OK but then closes immediately with `EBADF` (bad file descriptor). I can dig into this more later if it's helpful but wanted to let you know in case you have any idea what might be causing that. I reverted to stable 4.x for now.

PR #980, opened 2016-10-26
Remove unnecessary dependencies

mscdex:
This PR removes unnecessary dependencies, reducing current installation size by 1.6MB.

### DIFF
--- a/lib/parser-byte-length.js
+++ b/lib/parser-byte-length.js
@@ -1,6 +1,6 @@
 'use strict';
 var Transform = require('readable-stream').Transform;
-var inherits = require('inherits');
+var inherits = require('util').inherits;
 
 function ByteLengthParser(options) {
   if (!(this instanceof ByteLengthParser)) {
--- a/lib/parser-byte-length.js
+++ b/lib/parser-byte-length.js
@@ -1,5 +1,5 @@
 'use strict';
-var Transform = require('readable-stream').Transform;
+var Transform = require('stream').Transform;
 var inherits = require('util').inherits;
 
 function ByteLengthParser(options) {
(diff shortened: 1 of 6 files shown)


lib/parser-byte-length.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
-var Transform = require('readable-stream').Transform;
-var inherits = require('inherits');
+var Transform = require('stream').Transform;
```
reconbot: Can't get rid of `readable-stream` as it's the core of the `streams` branch and provides a common stream interface across all supported nodejs versions

lib/parser-delimiter.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
-var Transform = require('readable-stream').Transform;
-var inherits = require('inherits');
+var Transform = require('stream').Transform;
+var inherits = require('util').inherits;
```
reconbot: this one is tiny, I started using to aid when chrome serialport bindings become at thing, but I'm ok with ditching it at this point.

lib/serialport.js
```diff
@@ -55,6 +52,29 @@ var defaultSetFlags = {
   rts: true
 };
 
+function assign(target, defaults, user) {
```
reconbot: I'd still rather polyfill instead of replace `Object.assign` when it's available. I'd also break this out into a module. And this implementation isn't correct. if you want to steal the [`object.assign`'s](https://github.com/sindresorhus/object-assign/blob/master/index.js#L59-L84) code I'm ok with it.

mscdex: My focus wasn't to strictly emulate `object.assign()`, it was to serve the needs of this particular module.

mscdex:
It's not so much just about reducing the total footprint, but also reducing installation time by reducing the number of fetches to the registry.

Feel free to do whatever you want with this.

lib/parser-delimiter.js
reconbot: I'm +1 on this change

PR #1018, opened 2016-11-17
Adding WindowsBinding reading

munyirik:
@reconbot I took a stab at https://github.com/EmergingTechnologyAdvisors/node-serialport/issues/994  today. 
I've only done minimal testing but I wanted to open the PR early and get your feedback.

### DIFF
--- a/lib/bindings-win32.js
+++ b/lib/bindings-win32.js
@@ -19,7 +19,33 @@ function WindowsBinding(opt) {
   this.fd = null;
 };
 
-WindowsBinding.prototype.read = require('./read-unix');
+WindowsBinding.prototype.read = function (buffer, offset, length, cb) {
+  if (!Buffer.isBuffer(buffer)) {
+    throw new TypeError('"buffer" is not a Buffer');
+  }
+
+  if (typeof offset !== 'number') {
+    throw new TypeError('"offset" is not an integer');
+  }
+  
+  if (typeof length !== 'number') {
+    throw new TypeError('"length" is not an integer');
+  }
+
+  if (buffer.length < offset + length) {
+    throw new TypeError('buffer is too small');
+  }
+
+  if (typeof cb !== 'function') {
+    throw new TypeError('"cb" is not a function');
+  }
+
+  if (!this.isOpen) {
+    return asyncError(cb, new Error('Port is not open'));
+  }
+
+  binding.read(this.fd, buffer, offset, length, cb);
+};
 
 WindowsBinding.prototype.open = function(path, options, cb) {
   if (!path) {
--- a/lib/bindings-win32.js
+++ b/lib/bindings-win32.js
@@ -19,7 +19,7 @@ function WindowsBinding(opt) {
   this.fd = null;
 };
 
-WindowsBinding.prototype.read = function (buffer, offset, length, cb) {
+WindowsBinding.prototype.read = function(buffer, offset, length, cb) {
   if (!Buffer.isBuffer(buffer)) {
     throw new TypeError('"buffer" is not a Buffer');
   }
--- a/src/serialport.cpp
+++ b/src/serialport.cpp
@@ -336,6 +336,56 @@ void EIO_AfterWrite(uv_work_t* req) {
   delete queuedWrite;
 }
 
+NAN_METHOD(Read) {
+  // file descriptor
+  if (!info[0]->IsInt32()) {
+    Nan::ThrowTypeError("First argument must be an int");
+    return;
+  }
+  int fd = Nan::To<v8::Int32>(info[0]).ToLocalChecked()->Value();
+
+  // buffer
+  if (!info[1]->IsObject() || !node::Buffer::HasInstance(info[1])) {
+    Nan::ThrowTypeError("Second argument must be a buffer");
+    return;
+  }
+  v8::Local<v8::Object> buffer = info[1]->ToObject();
+  char* bufferData = node::Buffer::Data(buffer);
+  size_t bufferLength = node::Buffer::Length(buffer);
+
+  // offset
+  if (!info[2]->IsInt32()) {
+    Nan::ThrowTypeError("Third argument must be an int");
+    return;
+  }
+  int offset = Nan::To<v8::Int32>(info[2]).ToLocalChecked()->Value();
+  
+  // bytes to read
+  if (!info[3]->IsInt32()) {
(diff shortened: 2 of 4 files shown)


src/serialport.cpp
```diff
@@ -681,6 +731,7 @@ extern "C" {
     Nan::SetMethod(target, "open", Open);
     Nan::SetMethod(target, "update", Update);
     Nan::SetMethod(target, "write", Write);
+    Nan::SetMethod(target, "read", Read);
```
reconbot: Should we move all of this to serialport_win.cpp

src/serialport.h
```diff
@@ -26,6 +26,10 @@ NAN_METHOD(Write);
 void EIO_Write(uv_work_t* req);
 void EIO_AfterWrite(uv_work_t* req);
 
+NAN_METHOD(Read);
+void EIO_Read(uv_work_t* req);
```
reconbot: And this stuff inside an ifdef

src/serialport_win.cpp
```diff
@@ -311,6 +296,107 @@ void EIO_Write(uv_work_t* req) {
...
+  Nan::HandleScope scope;
+  bool skipCleanup = false;
+  ReadBaton* data = static_cast<ReadBaton*>(req->data);
+
+  v8::Local<v8::Value> argv[3];
+	
+  if (data->errorString[0]) {
+    argv[0] = Nan::Error(data->errorString);
+	argv[1] = argv[2] = Nan::Undefined();
+	Sleep(100); // prevent the errors from occurring too fast
```
reconbot: Is this left over from when we used to loop endlessly?

src/serialport_win.cpp
```diff
@@ -311,6 +296,107 @@ void EIO_Write(uv_work_t* req) {
...
+  } else {
+    argv[0] = Nan::Null();
+	argv[1] = Nan::New<v8::Integer>((int)data->bytesRead);
+	argv[2] = Nan::NewBuffer(data->bufferData, data->bufferLength, FinalizerCallback, req).ToLocalChecked();
```
reconbot: Yeah, we have a buffer from js land, this path seems not needed

src/serialport_win.cpp
munyirik: Yes. Is the sleep still needed? I didn't want to take it out in case it was there for a good reason.

src/serialport.cpp
munyirik: I'd like to leave it since serialport_uwp will also use it. I can use ifdef WIN32 since the macro is defined when UWP is used.

munyirik: or maybe I could keep it in a 'common' cpp file. I could avoid using an ifdef if there was something like a SetExtraMethods() in the init function where native bindings could add extra methods (in this case read) or just do nothing. What do you think?

reconbot: lets just `ifdef win32` but make sure to put the after read in this file too

munyirik:
You mean in EIO_AfterRead? There isn't any additional read. It's just for setting up calling the readcallback.

src/serialport.h
```diff
@@ -141,6 +145,17 @@ struct QueuedWrite {
   }
 };
 
+struct ReadBaton {
+  int fd;
+  char* bufferData;
```
reconbot: So I think the crux of the change will be to make this a pointer to a `v8::Local<v8::Object> buffer` object, that way we can pass it all the way through. Write to it and then pass it to the callback.  I don't know how to write to a buffer as if it were a `char[]` maybe @indutny can throw us some pointers.

munyirik:
@reconbot could you take a look at the latest updates when you get a chance.

lib/bindings-win32.js
```diff
@@ -19,7 +19,33 @@ function WindowsBinding(opt) {
   this.fd = null;
 };
 
-WindowsBinding.prototype.read = require('./read-unix');
+WindowsBinding.prototype.read = function(buffer, offset, length, cb) {
+  if (!Buffer.isBuffer(buffer)) {
+    throw new TypeError('"buffer" is not a Buffer');
```
reconbot: I'm pretty sure your c++ throws would bubble up just fine for type checking

src/serialport.cpp
```diff
@@ -336,6 +336,76 @@ void EIO_AfterWrite(uv_work_t* req) {
...
+  ReadBaton* data = static_cast<ReadBaton*>(req->data);
+
+  v8::Local<v8::Value> argv[2];
+    
+  if (data->errorString[0]) {
+    argv[0] = Nan::Error(data->errorString);
+    argv[1] = Nan::Undefined();
+  } else {
+    argv[0] = Nan::Null();
+    argv[1] = Nan::New<v8::Integer>((int)data->bytesRead);
```
reconbot: We should be calling the read callback with the [buffer as the 3rd arg](https://github.com/EmergingTechnologyAdvisors/node-serialport#module_serialport--SerialPort..readCallback). The decision was to stay in party with `fs.read`.

src/serialport_win.cpp
```diff
@@ -311,6 +283,72 @@ void EIO_Write(uv_work_t* req) {
...
+
+  while (true) {
+    OVERLAPPED ov = {0};
+    ov.hEvent = hEvent;
+
+    // Start read operation - synchrounous or asynchronous
+    DWORD bytesReadSync = 0;
+    if (!ReadFile((HANDLE)data->fd, offsetPtr, data->bytesToRead, &bytesReadSync, &ov)) {
+      errorCode = GetLastError();
+      if (errorCode != ERROR_IO_PENDING) {
```
reconbot: might be worth looking at https://github.com/EmergingTechnologyAdvisors/node-serialport/pull/1020

lib/bindings-win32.js
munyirik: Yeah seems like overkill

src/serialport.cpp
munyirik: Ok. I had left it out since the push-methods _read wasn't using that parameter - https://github.com/EmergingTechnologyAdvisors/node-serialport/blob/master/lib/push-methods.js#L31

src/serialport_win.cpp
munyirik: Thanks. I'll keep my eye on it. Looks like someone's still able to repro with the proposed fix.

Issue #877, opened 2016-07-20
It seems that on Windows OS vendorId always set to undefined


fduman:
Operating System and Hardware: Windows 10
NodeJS Version: 5.10.0
Serialport version: 2.1.2

I use serialport's list method and I cannot get vendorId information from that function. I had to use the pnpId property.

It seems that serialport_win.cpp has no code to get the vendor id.

I can get the vendor id information on OS X and this is an inconsistent behavior.

reconbot:
This might be a bug. 

 - We totally don't get them https://github.com/EmergingTechnologyAdvisors/node-serialport/blob/master/src/serialport_win.cpp#L493-L504
 - I have no idea if we can get them

We're using a library called http://github.com/todbot/usbSearch/ to find ports and get info. This area is ripe for cleaning them up.

Zensey:
Also there is no productId and locationId

PR #1019, opened 2016-11-20
Refactoring of EIO_List for Windows:  …

Zensey:
1. use single API -- SetupAPI directly without COM-object
2. supply ResultItem with vendorId, productId and locationId (issue #877)
3. solve problem with pnpId not beeing returned on some systems


Zensey:
thanks!

PR #1060, opened 2017-01-10
package.json: update node-pre-gyp config

### LINKED ISSUE #1055: Use `binary.remote_path` in package.json so host can be overridden
So this is an oddity in how `node-pre-gyp` works and furthermore my use case:  I'm trying to publish pre-built ARM binaries so I don't have to do native build on-target.

To accomplish this I can (mostly) just override flags during publish and install scripts.  Publish looks like this:
```bash
export node_pre_gyp_accessKeyId=xxx
export node_pre_gyp_secretAccessKey=xyz
export node_pre_gyp_bucket=mybucket

$(npm bin)/node-pre-gyp package publish
```

which publishes to `https://mybucket.s3.amazonaws.com/EmergingTechnologyAdvisors/node-serialport/releases/download/4.0.7/serialport-v4.0.7-node-v48-linux-arm.tar.gz` - great!

However when doing `npm install` the node-pre-gyp mirror location is looked up differently - it's specified as `opts.module_name + '_binary_host_mirror'` (command-line option) or `package_json.binary.host` and then [concatenates package_json.binary.remote_path](https://github.com/mapbox/node-pre-gyp/blob/master/lib/util/versioning.js#L301-L306).  

Because serialport defines the path as part of the host in package.json, my install needs to look like this:

```bash
npm i --production --serialport_binary_host_mirror=https://mybucket.s3.amazonaws.com/EmergingTechnologyAdvisors/node-serialport/releases/download/4.0.7/
``` 
...which is less than ideal versus just passing `--serialport_binary_host_mirror=https://mybucket.s3.amazonaws.com`

We can accomplish that if you split the [host line in package.json](https://github.com/EmergingTechnologyAdvisors/node-serialport/blob/master/package.json#L13) into `host` and `remote_path`. It's then possible to only override the host and keep a consistent path between publish and install steps.

Also because node-pre-gyp will interpolate those strings you can use a variable for the version portion of the path:
```
"binary: {
  "host": "https://github.com",
  "remote_path": "EmergingTechnologyAdvisors/node-serialport/releases/download/{version}"
}
```

this is similar to how sqlite3 defines their node-pre-gyp settings and I've confirmed what I'm trying to accomplish works for that module: https://github.com/mapbox/node-sqlite3/blob/master/package.json#L13-L14

I believe making this change for serialport should not have any adverse affect - future releases should look exactly the same otherwise.

thom-nic:
separate `host` and  `remote_path` portions so a mirror location can be provided for private prebuilt binaries.

Fixes #1055

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
   "binary": {
     "module_name": "serialport",
     "module_path": "build/{configuration}/",
-    "host": "https://github.com/EmergingTechnologyAdvisors/node-serialport/releases/download/4.0.7"
+    "host": "https://github.com",
+    "remote_path": "/EmergingTechnologyAdvisors/node-serialport/releases/download/{version}"
   },
   "main": "./lib/serialport",
   "repository": {


thom-nic:
This worked for me - but I would try running a build that publishes prebuilt binaries just to sanity check that nothing has changed for you.  I'll branch and make a PR for 5x next.

PR #1063, opened 2017-01-12
Revert "package.json: udpate node-pre-gyp params host & remote_path for v5x"

reconbot:
Reverts EmergingTechnologyAdvisors/node-serialport#1061


thom-nic:
@reconbot did you find some issue with this?

thom-nic:
Ah thanks sorry I totally missed that.  Will investigate.

Issue #1188, opened 2017-06-10
Write failure after a number of commands


techninja:
- __SerialPort Version__: v5 beta3 tag
- __NodeJS Version__: node 6, electron 1.6.10
- __Operating System__ and __Hardware Platform__: Win 10, x64 (possibly other win)

- Have you checked the right __version__ of the __api docs?__: 👍 
- Are you using __Electron__ and have you checked the _Electron Docs_?: Yup. Fails inside and outside electron

### Summary of Problem
While doing extensive testing for the latest version of [my Electron based software for streaming serial data to drawing robots](https://github.com/evil-mad/robopaint), I find that on Windows only, my data write callback loop to move through my array buffered command set will suddenly... __stop cold__. Somewhere around 3k-5k commands in. Debugging the process in Electron (running inside its own dedicated webview process) shows that the last serial write is triggered, but the write callback is not. Implementing a timeout to catch this does work, but re-initiating when this happens.. does not. Running flush seems to get the last command to write (the robot moves), but running writes after this does not work.

### Steps and Code to Reproduce the Issue

I do not have a good test case apart from my description above. If prompted I could likely try to make a short form test for reproduction, but without understanding the problem more, the best I can say is that it appears to very low level, not a problem with the destination hardware, and ___probably___ related to the number of commands sent. Maybe call stack limit? Who knows.

techninja:
I added an explicit counter to see exactly how many commands were being sent+drained when this happened. Ran 5 tests. The first two had exactly the same circumstances, and died in the exactly the same way. The fourth I reset the bot position full manual, and ended up with more commands before hitting the bug. The last two were completely different art at a much lower, then much higher speed.

__Test results:__
1. Firefox @ 35% speed, Restart (15), then reset position (19), dies at (3067)
2. Firefox @ 35% speed, Restart (15), then reset position (19), dies at (3067)
3. Firefox @ 35% speed, Restart (15), dies at 3074
4. Bird at 15% speed, Restart (15, 19), dies at (3011)
5. Bird at 75% speed Restart (15), dies at (3018)

I take from this:
* Drain/send speed doesn't seem to effect it (each movement command has a timer on it that restricts drain event until after execution on the firmware, so for faster speeds, movements avg ~1-10ms instead of 50-100ms)
* The number of commands is definitely a factor, and a lot less than I thought
* The loop that calls this may be a factor, so will retest with a rethinking of how this works

techninja:
[...]

PR #1212, opened 2017-06-27
[unix] New read/write subsystem

reconbot:
This has always been the dream. Use `fs.read` and `fs.write` to do reading and writing. To make that happen I've created a cross platform `Poller` object that notifies us once on `readable` and `writable` states.

- Reduce large write cpu usage from 100% to 1-2%
- Replace ReadPoller with a multi purpose poller because you can only have a single uv_poll_t on a fd at a time because reasons
- Experimented with doing this all in c++, maybe there are gains to doing that but I'd never finish.
- Some unrelated eslint changes
- Arduino echo should just echo, writing random bytes was messing it up
- use a file to disable tests in envs
- read and write errors can trigger disconnects
- disconnect event only happens if there are listeners, otherwise error events will be triggered
- a disconnect during write will always have a write error too
- win32 bindings seem to only trigger disconnects on writes

Todo
- [x] Remove the excessive logging
- [x] Remove *nix read/write c++ methods and ifdef out the shared code
- [x] Test poller on windows (didn't work)
- [x] use `uv_close` in the poller to free the handle's memory
- [x] cleanup the poller api
- [x] test disconnecting and closing during reads, writes, and polling, clearer ways to deal with this now.

works #1188


CLEMARCx:
I have a Windows dev environment running, so I can test it if you want.

PR #1222, opened 2017-07-08
Mocking examples and testing entry point

### LINKED ISSUE #1066: Provide an example for testing with mock bindings
SerialPort version: 5.0.0

Maybe also provide some docs on the mock binding's object's options.

reconbot:
closes #1066

### DIFF
--- a/.docs/README.hbs
+++ b/.docs/README.hbs
@@ -59,6 +59,7 @@ In addition to reading the [article mentioned above](http://www.voodootikigod.co
   * [Windows](#windows)
 * [Usage](#usage)
   * [Opening a Port](#opening-a-port)
+  * [Testing](#testing)
   * [Debugging](#debugging)
   * [Error Handling](#error-handling)
 {{#class name="SerialPort"}}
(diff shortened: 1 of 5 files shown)


vkuehn:
that is so cool ! you may want to mention it on the front readme..
However I think ./example/mockeing.js does not work with that requiere path, should be 
const SerialPort = require('../test'); ?

vkuehn:
oh ok just did download node-serialport and a node .\examples\mocking.js which didn't work. How would I install that Version into a regular app  without a lot of handish work ?

vkuehn:
did install with npm install serialport@5.0.0-beta7 --save
still getting this for mocking.js
port opened
Write callback returned
Last write:	 undefined

the only thing I changed is "() =>" to "function ()"
node version is v6.10.3 would I need a newer version ?

vkuehn:
sure it is basically the example file

renamed because of githubs upload constraints
[mocking.txt](https://github.com/EmergingTechnologyAdvisors/node-serialport/files/1139181/mocking.txt)

vkuehn:
ok will do some testing on other machines and than open the issue

PR #1248, opened 2017-07-22
call udevadm single time

akaJes:
your version of linux-list calls numerous processes of the shell and the udevadm
this PR needs the byline module, but calls the udevadm single time
it most suitable for serial ports polling


akaJes:
the list-linux becomes lighter in twice :)
and the last travis check has suspicious internal error

i'm tired to fight with lint, test and travis :)

Issue #1229, opened 2017-07-13
Data clipping over 64 bytes


tuna-f1sh:
- __SerialPort Version__: 5.0.0-beta7
- __NodeJS Version__: 8.1.3
- __Operating System__ and __Hardware Platform__: macOS, wchusbserial USB serial 

- Have you checked the right __version__ of the __api docs?__: YES
- Are you having trouble installing and you checked the __Installation Special Cases__ docs? NO
- Are you using __Electron__ and have you checked the _Electron Docs_?: NO

### Summary of Problem

My data stream seems to be being clipped to around 64 bytes. I'm trying to send 120 bytes in total but they don't come through. This sounds like a noddy issue but it's really got me stumped!

I'm driving a Hanover Flipdot display over RS485. I have a driver in Python working successfully but wish to use Node for an app idea I have. I've used nodeserial in the past successfully but not with large buffers.

### Steps and Code to Reproduce the Issue

The code I'm developing here: https://github.com/tuna-f1sh/flipdot

I can reproduce it however simply by sending a buffer larger than 64 bytes:

```
var SerialPort = require('serialport');
var port = new SerialPort('/dev/tty.wchusbserial1420', {
  baudRate: 4800
});

port.write(new Buffer(256), function(err) {
  if (err) {
    return console.log('Error on write: ', err.message);
  }
  console.log('message written');
});

// Open errors will be emitted as an error event
port.on('error', function(err) {
  console.log('Error: ', err.message);
});
```
![screen shot 2017-07-13 at 13 31 48](https://user-images.githubusercontent.com/1886746/28167872-16327530-67d5-11e7-9737-aeaa7d85363f.png)

Diagnosed by scoping the output. You'll notice the baud rate is quite slow (4800) but should take 533 ms to transmit 256 bytes. The message is ~133 ms then ends however (64 bytes). Changing the baud rate (doubling to 9600), only halves the time ~66 ms so it still clips. 

Shortening the buffer to 1 byte and running the script above, the transmission will remain 64 bytes for a few runs before just transmitting 1 byte as requested - it's although the driver is unloading the previous data at write!...

I've confirmed that the `highWaterMark` is nothing silly and even set it to 32 kB. I've tried a lot of things so hopefully this isn't something stupid and is a bug?

reconbot:
[...]

PR #1256, opened 2017-07-26
Write/Drain and Readback Manual Test

tuna-f1sh:
Following #1229, I've added two files to manual-testing to verify the operation of `write()` and in particular, `drain()` events.

The the uses two USB serial devices to form a loopback test. I've used an additional environment variable `TEST_PORT_RX` to denote the receiver. Example usage:

`TEST_PORT=/dev/ttyUSB0 TEST_PORT_RX=/dev/ttyUSB1 DEBUG=* node read.js`

TX from `TEST_PORT` should be connected to RX on `TEST_PORT_RX`.

It will exit with ok (0) if number of bytes received matches number of bytes expected, otherwise error (1). Console log messages show sent data length and received data length. The script should be called from the script directory due to the child process calling the drain.js script explicitly - might be worth changing this to pass CWD.

### DIFF
--- a/test/manual-testing/read.js
+++ b/test/manual-testing/read.js
@@ -1,0 +1,40 @@
+/* eslint-disable node/no-missing-require */
+'use strict';
+const SerialPort = require('../../');
+const port = process.env.TEST_PORT_RX;
+const { exec } = require('child_process');
+
+const expected = 512;
+
+if (!port) {
+  console.error('Please pass TEST_PORT environment variable');
+  process.exit(1);
+}
+
+const serialPort = new SerialPort(port, (err) => {
+  if (err) { throw err }
+  exec('node drain.js', (err, stdout) => {
+    if (err) {
+      // node couldn't execute the command
+      process.exit(1);
+    }
+
+    console.log(stdout);
+
+    setTimeout(() => {
+      serialPort.on('data', (data) => {
+        console.log(`Recieved data dength: ${data.length} B`);
+        if (data.length === expected) {
+          process.exit(0);
+        } else {
+          process.exit(1);
+        }
+      });
+    }, 100);
+
+    setTimeout(() => {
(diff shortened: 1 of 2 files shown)


test/manual-testing/read.js
```diff
@@ -0,0 +1,47 @@
...
+const SerialPort = require('../../');
+const { exec } = require('child_process');
+
+// Serial receiver device 
+const port = process.env.TEST_PORT_RX;
+// Expected number of bytes to receive (should make `size` in drain.js)
+const expected = 512;
+
+if (!port) {
+  console.error('Please pass TEST_PORT environment variable');
```
reconbot: Should be `TEST_PORT_RX`

test/manual-testing/read.js
```diff
@@ -0,0 +1,47 @@
...
+  exec('node drain.js', (err, stdout) => {
+    if (err) {
+      // node couldn't execute the command
+      process.exit(1);
+    }
+
+    console.log(stdout);
+
+    // Read back the data received on the read device after a short timout to ensure transmission
+    setTimeout(() => {
```
reconbot: I'd use the `byteLength` parser to get a single data event and start the read right away, no need to mess with timing on reads.

test/manual-testing/read.js
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable node/no-missing-require */
+'use strict';
+const SerialPort = require('../../');
+const { exec } = require('child_process');
```
reconbot: We're still supporting node4 so this has to be `const exec = require('child_process').exec;`

tuna-f1sh:
@reconbot Thanks. I've taken into account the review - learnt something new with the `parser`. All lints fine now too.  Not sure if the coverage is something I can resolve?

PR #1261, opened 2017-07-29
[all] Change parsers to be es6 classes

reconbot:
- Reduces code a bit
- Provides a better code example
- clean up options passed into transform streams

On @frank-dspeed's prompting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -723,52 +723,62 @@ The default `Parsers` are [Transform streams](https://nodejs.org/api/stream.html
 
 **Example**  
 ```js
-var SerialPort = require('serialport');
-var Readline = SerialPort.parsers.Readline;
-var port = new SerialPort('/dev/tty-usbserial1');
-var parser = new Readline();
+const SerialPort = require('serialport');
+const Readline = SerialPort.parsers.Readline;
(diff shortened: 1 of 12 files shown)


frank-dspeed:
@reconbot wait with that pull i have something special for you
to also apply as the tests are also designed to take the constructors without new i created a decorator for that!!

frank-dspeed:
@reconbot 
https://github.com/direktspeed/class-nonew-decorator

look into that!

```
const nonew = require('class-nonew-decorator');

@nonew()
class IamClass {
    constructor() {
        console.log('IamClass Instance!');
    }
}

console.log(new IamClass() instanceof IamClass);  // true 

console.log(IamClass() instanceof IamClass);  // true
```

you require it as const and add it with @ over the class it will wrap the class app so it works without new thats normaly not possible its tricky :D

frank-dspeed:
@reconbot designed it only to make the new cctalk parser for you and still allow the usage without new as class :+1:

frank-dspeed:
@reconbot we can use it no problem i refactor your lib to be ES6 Compatible via bable if you like

frank-dspeed:
@reconbot #1262 - added es7 including nice build :+1:

PR #1262, opened 2017-07-29
add_es7_build

frank-dspeed:


frank-dspeed:
Ok Some infos about that
- we npmignore now lib/ and made dist/ master 
- dist gets auto created via pre publish hook
- dist gets created on test
- added also to mocha tests but we should refactor tests to also use the dist/

frank-dspeed:
ok tests are now using also dist happy merging no discussion needed probally :+1: 
it lets us use lib like befor and creates a nice dist for npm

frank-dspeed:
@reconbot review please ;)

test/bindings.js
```diff
@@ -51,7 +51,7 @@ const readyData = Buffer.from('READY');
 
 // Test our mock binding and the binding for the platform we're running on
 bindingsToTest.forEach((bindingName) => {
-  const Binding = require(`../lib/bindings/${bindingName}`);
+  const Binding = require(`../dist/bindings/${bindingName}`);
```
reconbot: Why point to `dist` if we're using `babel/register`?

.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "presets": [
+    "es2015",
+    "stage-0"
```
reconbot: I'd rather not use stage 0 features, stage 4 has plenty of the missing pieces from node. Unless there's something specific.

package.json
```diff
@@ -62,6 +62,7 @@
     }
   ],
   "dependencies": {
+    "babel-polyfill": "^6.23.0",
```
frank-dspeed: and i aggree this don't needs a pollyfill i think as we don't load it :) and build directly

frank-dspeed: :+1:

test/bindings.js
frank-dspeed: don't know maybe done that in a hurry your right the tests probally don't need to use dist :+1:

.babelrc
frank-dspeed: @reconbot it gets only used for dev it has nothing to do with the production build stage-0 means we can simply use all stuff it makes no sense for dev to add them indipendent dev overhead is nothing we should care of?

frank-dspeed: :-1:

frank-dspeed:
@reconbot if your fast you can cancel that build and i remove the test to dist upgrade and also the pollyfill as we don't load it and build befor. :+1:

package.json
frank-dspeed: Even when we use babel-register the poly is needed so we can move it to devDependencys :+1: 

>Polyfill not included
You must include the polyfill separately when using features that require it, like generators.

frank-dspeed:
@reconbot ok

PR #1313, opened 2017-08-25
Asynchronous callbacks for reading and writing on Windows. Fixes #1221

### LINKED ISSUE #1221: [windows] Use non blocking reads and writes
*We need help with this one!*

- __SerialPort Version__: 5.0.0-beta6
- __Operating System__ and __Hardware Platform__: Windows / All

Right now we can only support [3 or 4 ports open at a time](https://github.com/EmergingTechnologyAdvisors/node-serialport/issues/914) in the same process on windows unless the env var `UV_THREADPOOL_SIZE` is set to something higher than 4. This is because we're using `WriteFile` and `ReadFile` for our reads and writes which are blocking. We should transition to [`WriteFileEx`](https://msdn.microsoft.com/en-us/library/aa365748(v=VS.85).aspx) and [`ReadFileEx`](https://msdn.microsoft.com/en-us/library/aa365468(v=VS.85).aspx) which support callbacks when data is read or written. This should allow us to support _"any"_ number of ports and have lower cpu utilization.

Relevant code; [EIO_Write](https://github.com/EmergingTechnologyAdvisors/node-serialport/blob/master/src/serialport_win.cpp#L301-L341) and [EIO_Read](https://github.com/EmergingTechnologyAdvisors/node-serialport/blob/master/src/serialport_win.cpp#L413-L477)

dustmop:
Instead of ReadFile and WriteFile, which block and transfer data synchronously, use ReadFileEx and WriteFileEx, which both allow async callbacks. In addition, change how timeouts are used for ReadFile*, using an unlimited timeout for the first byte, and no timeout for the rest of the data in the input buffer. This removes the need to poll entirely, while still retrieving all data available in the input buffer. In both cases, the I/O operations happen in their own threads, since Windows requires IOCompletion callbacks to wait for their calling thread to be in an "alertable wait state".

Fixes #1221

### DIFF
--- a/src/serialport_win.cpp
+++ b/src/serialport_win.cpp
@@ -14,14 +14,7 @@
 // Declare type of pointer to CancelIoEx function
 typedef BOOL (WINAPI *CancelIoExType)(HANDLE hFile, LPOVERLAPPED lpOverlapped);
 
-void EIO_AfterWrite(uv_async_t* req);
-DWORD WriteThread(void* param);
-
-void EIO_AfterRead(uv_async_t* req);
-DWORD ReadThread(void* param);
-
 std::list<int> g_closingHandles;
-COMMTIMEOUTS g_commTimeouts = {0};
 
 
 void ErrorCodeToString(const char* prefix, int errorCode, char *errorStr) {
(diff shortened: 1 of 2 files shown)


src/serialport_win.cpp
```diff
@@ -14,7 +14,15 @@
 // Declare type of pointer to CancelIoEx function
 typedef BOOL (WINAPI *CancelIoExType)(HANDLE hFile, LPOVERLAPPED lpOverlapped);
 
+void EIO_AfterWrite(uv_async_t* req);
```
reconbot: pretty sure this can go in `serialport_win.h`

src/serialport_win.cpp
```diff
@@ -14,7 +14,15 @@
...
+DWORD WriteThread(void* param);
+
+void EIO_AfterRead(uv_async_t* req);
+DWORD ReadThread(void* param);
+
 std::list<int> g_closingHandles;
+COMMTIMEOUTS g_commTimeouts = {0};
```
reconbot: Will this be shared between all open serialports? If so, would that cause trouble as the timeouts are adjusted?

src/serialport_win.cpp
dustmop: Sure thing, moved.

src/serialport_win.cpp
dustmop: I hadn't thought about multiple devices using a single library instance. You are correct, this shouldn't be a global.

munyirik:
LGTM. Unfortunately I can't test.

noopkat:
sorry I'm so late to the testing party 😞 can I still help with this?

dustmop:
I think I can take a look at fixing this tomorrow. What should I run to see if the full test suite is passing?

noopkat:
@dustmop if you bump this PR again when it's ready I'll run the tests with my arduino too on Win 10

dustmop:
@noopkat New pull request #1328.

PR #1328, opened 2017-09-06
Asynchronous callbacks for reading and writing on Windows, attempt 2.

### LINKED ISSUE #1221: [windows] Use non blocking reads and writes

dustmop:
Instead of ReadFile and WriteFile, which block and transfer data synchronously, use ReadFileEx and WriteFileEx, which both allow async callbacks. In addition, change how timeouts are used for ReadFile*, using an unlimited timeout for the first byte, and no timeout for the rest of the data in the input buffer. This removes the need to poll entirely, while still retrieving all data available in the input buffer. In both cases, the I/O operations happen in their own threads, since Windows requires IOCompletion callbacks to wait for their calling thread to be in an "alertable wait state".

Fixes #1221

### DIFF
--- a/src/serialport_win.cpp
+++ b/src/serialport_win.cpp
@@ -89,8 +89,8 @@ void EIO_Open(uv_work_t* req) {
   dcb.Parity = NOPARITY;
   dcb.ByteSize = 8;
   dcb.StopBits = ONESTOPBIT;
-  
-  
+
+
   dcb.fOutxDsrFlow = FALSE;
   dcb.fOutxCtsFlow = FALSE;
 
@@ -152,11 +152,10 @@ void EIO_Open(uv_work_t* req) {
     return;
   }
 
-  // Set the timeouts for read and write operations read operation is to return
-  // immediately with the bytes that have already been received, even if no bytes
-  // have been received.
-  COMMTIMEOUTS commTimeouts = {0};
-  commTimeouts.ReadIntervalTimeout = MAXDWORD;  // Never timeout
+  // Set the timeouts for read and write operations.
+  // Read operation will wait for at least 1 byte to be received.
+  COMMTIMEOUTS commTimeouts = {};
+  commTimeouts.ReadIntervalTimeout = 0;         // Never timeout, always wait for data.
   commTimeouts.ReadTotalTimeoutMultiplier = 0;  // Do not allow big read timeout when big read buffer used
   commTimeouts.ReadTotalTimeoutConstant = 0;    // Total read timeout (period of read loop)
   commTimeouts.WriteTotalTimeoutConstant = 0;   // Const part of write timeout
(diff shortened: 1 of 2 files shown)


dustmop:
Verified this builds in 32-bit mode using:
`node-gyp rebuild --msvs_version=2015 --arch=ia32`

Ran tests using
`TEST_PORT=COM3 npm test`
and got:
`  233 passing (17s)
  3 pending`

Looks like the 3 pending tests are features that are disabled on Windows?

noopkat:
👋  is this ready for me to test on my Windows box?

noopkat:
running this on 64bit Surface Book gives me the following:

```
     #read
        √ errors asynchronously when not open
        1) doesn't throw if the port is open


  79 passing (7s)
  2 pending
  1 failing

  1) bindings/win32 instance method #read doesn't throw if the port is open:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

noopkat:
@reconbot it is consistent. I tried board resetting, repeated testing, and ran tests in both git-bash and powershell with the same results each time.

dustmop:
On my machine:

```
bindings/win32
...
    instance method
...
      #read
        √ errors asynchronously when not open
        √ doesn't throw if the port is open (1582ms)
        √ returns at maximum the requested number of bytes (1579ms)```

1582ms is pretty close-ish to 2000ms? Is there a way to try with an increased timeout?

**Comment by noopkat** [comment]
> 1582ms is pretty close-ish to 2000ms? Is there a way to try with an increased timeout?
this machine is a good example of a 🐢 to test with, if that provides context around this at all

**Comment by noopkat** [comment]
`TEST_PORT=COM6 npx mocha --no-timeouts` = 🆗 👌

**Comment by dustmop** [comment]
> However that test does seem really slow, all it's doing is opening the port and waiting for the first byte

Small correction: this test (doesn't throw if the port is open) reads multiple bytes, equal to readyData.length. It's the test below (returns at maximum the requested number of bytes) that only reads a single byte. However, I agree it still seems to be on the slow side.

I want to make sure this isn't a regression. Was an increased timeout needed before this change? Also, is there a benchmark, or at least something that validates that cpu usage goes down? It seems good on my machine but would be nice to verify.

**Comment by noopkat** [comment]
weirdly trying from a master build will yield 

```
   optional locking
          √ locks the port by default
          1) can unlock the port


  54 passing (372ms)
  2 pending
  1 failing

  1) bindings/win32 instance method #open optional locking can unlock the port:
     Error: Opening COM6: Access denied
      at Error (native)
```

I do have to 😴 right now my apologies - early day in the city tomorrow. Will bring a board with me and can continue to debug if needed.

**Comment by noopkat** [comment]
so this dummy here did not pull master before building (it is late) 🤦‍♀️ 

I get the exact same `doesn't throw if the port is open` as earlier tonight. Doesn't seem like you caused a regression, @dustmop ?

**Comment by dustmop** [comment]
Ok! Thank you very much for helping, go get some rest! 👍

**Comment by dustmop** [comment]
Sure! Maybe wait until tomorrow, just in case?

**Comment by dustmop** [comment]
Looks like this should be good to go

**Comment by vshymanskyy** [comment]
Hi all. I had a chance to try it out on our real-life application (Electron, Windows, x86 build).
It uses Modbus to communicate with a device. I use SerialPort v6.0.0-beta2.
It starts OK, I can connect to my device and exchange some packets. But after 5-10 seconds it seems to stop working. I'm investigating further

**Comment by vshymanskyy** [comment]
OK, this is what happens:
```
2017-09-07T10:12:34.021Z serialport:bindings write 8 bytes
2017-09-07T10:12:34.029Z serialport:main binding.write error Error: Writing to COM port (WriteFileEx): Unknown error code 87
2017-09-07T10:12:34.029Z serialport:main disconnected Error: Writing to COM port (WriteFileEx): Unknown error code 87
2017-09-07T10:12:34.029Z serialport:main #close
2017-09-07T10:12:34.033Z serialport:bindings close
2017-09-07T10:12:34.037Z modbus-serial Error:┬á Writing to COM port (WriteFileEx): Unknown error code 87
2017-09-07T10:12:34.037Z serialport:main binding.read error { Error: Reading from COM port (ReadIOCompletion): Operation aborted canceled: true } 2017-09-07T10:12:34.041Z serialport:main _read queueing _read for after open
2017-09-07T10:12:34.155Z serialport:main binding.close finished
2017-09-07T10:12:34.155Z modbus-serial Close:┬á { Error: Writing to COM port (WriteFileEx): Unknown error code 87 disconnected: true }
```

dustmop:
Error code 87 is Windows' very unhelpful "ERROR_INVALID_PARAMETER", which doesn't say much. Could be the file handle is being invalidated, or the buffer is, somehow. Hard to figure out without a minimal repro case. This is 32-bit or 64-bit?

I haven't done any testing in electron, and won't be able to until I'm back home from travel (September 20th).

vshymanskyy:
Works fine with the latest version. Thank you!

PR #1342, opened 2017-09-22
Add cctalk parsers

frank-dspeed:
this adds the discussed cctalk parser
https://github.com/EmergingTechnologyAdvisors/node-serialport/issues/1240

### DIFF
--- a/lib/parsers/cctalk.js
+++ b/lib/parsers/cctalk.js
@@ -11,7 +11,7 @@ module.exports = class CCTalkParser extends Transform {
     this.cursor += buffer.length;
     // TODO: Better Faster es7 no supported by node 4
     const array = Array.prototype.slice.call(buffer, 0); // [...buffer];
-    array.map((byte) => this.array.push(byte));
+    array.map((byte) => this.array.push(byte)); // ES7 allows directly push
 
     // full frame accumulated
     while (this.cursor > 1 && this.cursor >= this.array[1] + 5) {
--- a/lib/parsers/cctalk.js
+++ b/lib/parsers/cctalk.js
@@ -12,7 +12,8 @@ module.exports = class CCTalkParser extends Transform {
     this.cursor += buffer.length;
     // TODO: Better Faster es7 no supported by node 4
     // ES7 allows directly push [...buffer]
-    buffer.map((byte) => this.array.push(byte));
+    Buffer.from(buffer)
+      .map((byte) => this.array.push(byte));
 
     while (this.cursor > 1 && this.cursor >= this.array[1] + 5) {
       // full frame accumulated
--- a/lib/parsers/cctalk.js
+++ b/lib/parsers/cctalk.js
@@ -1,5 +1,6 @@
 'use strict';
 const Transform = require('stream').Transform;
+const Buffer = require('safe-buffer').Buffer;
 
 module.exports = class CCTalkParser extends Transform {
   constructor() {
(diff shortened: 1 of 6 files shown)


frank-dspeed:
This is by the way 100% tested on real hardware in production its based on work:
https://github.com/direktspeed/serialport-parsers-cctalk

Only made compatible because i use some extra dependencies because of the ES7 and will switch to debug-extra https://github.com/direktspeed/debug-extra
that i created to debug streams and that stuff
its like fs extra extending debug with some stream and promise stuff

lib/parsers/cctalk.js
```diff
@@ -0,0 +1,39 @@
...
+
+module.exports = class ccTalkParser extends Transform {
+  constructor() {
+    super();
+    this.buffer = Buffer.alloc(MAX_PACKET_LENGTH); // maxium ccTalkMessage length
+    this.cursor = 0;
+  }
+  _transform(buffer, _, cb) {
+    debug('parser set')(this.buffer, this.cursor);
+    debug('parser set')(this.buffer.toString('hex'), buffer.buffer, this.cursor);
```
reconbot: I'd either push this into a conditional `debug.enabled` or use a [custom formatter](https://www.npmjs.com/package/debug#custom-formatters) so `toString` doesn't get called when not debugging.

frank-dspeed:
@reconbot can you please contribute to the tests?

we need to test the following Arrays

write [2,0,1,254,217] // its length 5 
NOTE: the secund byte 0 indicates that the length is 5+0=5
answer [1,0,2,0,217]

write [2,2,1,254,1,1,217] // Length 7 
NOTE: the secund 2 indicates that the length is 5+2= 7
answer [1,0,2,0,217]

frank-dspeed:
@reconbot and please look into the current travis error i don't know any more whats going on as 
it complains about buffer set wich is not really the same like buffer = that is now deprecated.

i don't know why they deprecated that because it makes sense to set some arrays at array position.

frank-dspeed:
@reconbot yes the tests are not done when i remove them and do it localy i get totall mess :)
```
     .isOpen
        ✓ is a read only property
        ✓ returns false when the port is created
        ✓ returns false when the port is opening
        1) returns false when the port is opening


  151 passing (237ms)
  11 pending
  1 failing

  1) SerialPort property .isOpen returns false when the port is opening:
     TypeError: Cannot read property 'then' of undefined
      at SerialPort.open (lib/serialport.js:232:46)
      at Context.it (test/serialport.js:238:14)
```

lib/parsers/index.js
```diff
@@ -76,6 +76,7 @@ parser.on('data', console.log);
```
reconbot: Add some docs up here ☝️

package.json
```diff
@@ -71,7 +71,7 @@
     "chai": "^4.0.2",
     "chai-subset": "^1.5.0",
     "conventional-changelog-cli": "^1.3.2",
-    "eslint": "^4.5.0",
+    "eslint": "^4.7.2",
```
reconbot: This is so weird, we totally have this in master

lib/parsers/cctalk.js
```diff
@@ -0,0 +1,33 @@
...
+const Transform = require('stream').Transform;
+
+module.exports = class CCTalkParser extends Transform {
+  constructor() {
+    super();
+    this.array = [];
+  }
+  _transform(buffer, _, cb) {
+    // TODO: Better Faster es7 no supported by node 4
+    const array = Array.prototype.slice.call(buffer, 0); // [...buffer];
```
reconbot:
```js
Array.from(buffer)
```
but this seems weird that casting from a buffer to an array to a `Uint8Array` is the right way forward.

frank-dspeed:
@reconbot i think you got it wrong because the answer should come from the device not from the parser the parser will simply emit the complet message written to it
so when we don't use your magic mockup framework to emit on that write the correct answer it can't work

frank-dspeed:
[2, 0, 1, 254, 217] writen to the mockup device should let the mock up device write [1, 0, 2, 0, 217]

for the parser its enought if he emits 2 commands from this write [2, 0, 1, 254, 217,2,2,1,254,1,1,217]

frank-dspeed:
@reconbot for the parser its enought if he emits 2 commands from this write [2, 0, 1, 254, 217,2,2,1,254,1,1,217]
so we need to check if he emits 2 times first time 2, 0, 1, 254, 217 secund time 2,2,1,254,1,1,217 thats it

frank-dspeed:
@reconbot but i don't understand it i will write a more good test 
```
var c = 0
parser.write(Buffer.from([2, 0, 1, 254, 217,2,2,1,254,1,1,217])
parser.on('data',(data)=>{
if (data.length === 5 || data.length === 7) {
  c++
if (c === 2) test passed
} else {
console.log('NOT GOOD')
}

})
```

lib/parsers/cctalk.js
frank-dspeed: this is because buffers don't support the needed methods they are really static we need a event chain so i choosed to go over to array

frank-dspeed:
@reconbot anything left undone ? it looks good to me?

PR #1367, opened 2017-10-18
Fix incorrect event loop behavior on Windows

kimushu:
See #1363

This patch solves two problems on Windows.

Bug#1. Event loop stops even if we have pending writes (and reads).
- Moved uv_async_init from WriteThread to NAN_METHOD(Write). This async handle will keep event loop alive.

Bug#2. Event loop does not stop even if all pending writes (and reads) has finished.
- Added uv_close to EIO_AfterWrite. This removes async handle from event loop.

### DIFF
--- a/src/serialport_win.cpp
+++ b/src/serialport_win.cpp
@@ -39,6 +39,12 @@ void ErrorCodeToString(const char* prefix, int errorCode, char *errorStr) {
   }
 }
 
+void AsyncCloseCallback(uv_handle_t* handle)
+{
+  uv_async_t* async = (uv_async_t*)handle;
+  delete async;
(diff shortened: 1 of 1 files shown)


src/serialport_win.cpp
```diff
@@ -293,9 +299,13 @@ NAN_METHOD(Write) {
   baton->offset = 0;
   baton->callback.Reset(info[2].As<v8::Function>());
   baton->complete = false;
+
+  uv_async_t* async = new uv_async_t;
```
reconbot: 👍

dustmop:
LGTM, thanks for the fixes!

PR #1388, opened 2017-11-03
Support 250k

### LINKED ISSUE #1362: Serialport communication fails at 250000 baudrate on Raspberry Pi
- __SerialPort Version__: 6
- __NodeJS Version__: 8.7
- __Operating System__ and __Hardware Platform__: Raspbian Jessie Lite on Raspberry pi

- Have you checked the right __version__ of the __api docs?__: Yes
- Are you having trouble installing and you checked the __Installation Special Cases__ docs? N/A
- Are you using __Electron__ and have you checked the _Electron Docs_?: N/A

### Summary of Problem
Unable to communicate with Arduino at 250000 baudrate. Identical code runs great on 2014 Mac running OSX 10.12
I've created a repo for reproducing this bug:
https://github.com/matthova/serialport-6-250000-test

matthova:
Adds support for custom baudrates for Raspberry Pi.

Closes #1362

### DIFF
--- a/src/custom_baudrate.cpp
+++ b/src/custom_baudrate.cpp
@@ -1,0 +1,14 @@
+#include <asm/termios.h>
+
+int ioctl(int fd, int request, ...);
+
+void set_custom_baudrate(const int fd, const int baudrate) {
+  #if defined(__linux__)
+    struct termios2 t;
+    ioctl(fd, TCGETS2, &t);
+    t.c_cflag &= ~CBAUD;
+    t.c_cflag |= BOTHER;
+    t.c_ospeed = t.c_ispeed = baudrate;
+    ioctl(fd, TCSETS2, &t);
+  #endif
+}
--- a/src/serialport_unix.cpp
+++ b/src/serialport_unix.cpp
@@ -24,6 +24,7 @@
 #if defined(__linux__)
 #include <sys/ioctl.h>
 #include <linux/serial.h>
+#include "custom_baudrate.h"
 #endif
 
 int ToStopBitsConstant(SerialPortStopBits stopBits);
@@ -100,6 +101,8 @@ int setBaudRate(ConnectionOptionsBaton *data) {
   // lookup the standard baudrates from the table
   int baudRate = ToBaudConstant(data->baudRate);
   int fd = data->fd;
+
+  set_custom_baudrate(fd, 250000);
 
   // get port options
   struct termios options;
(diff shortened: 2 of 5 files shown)


src/serialport_unix.cpp
```diff
@@ -101,6 +102,8 @@ int setBaudRate(ConnectionOptionsBaton *data) {
   int baudRate = ToBaudConstant(data->baudRate);
   int fd = data->fd;
 
+  set_custom_baudrate(fd, 250000);
```
matthova: Will need to modify hardcoded baudrate.

src/custom_baudrate.cpp
```diff
@@ -0,0 +1,14 @@
+#include <asm/termios.h>
+
+int ioctl(int fd, int request, ...);
+
+void set_custom_baudrate(const int fd, const int baudrate) {
```
reconbot: can this go into the existing set buadrate function?

matthova: I'm not too sharp on C++, I had a friend help make the code compile. In order to have it recognize termios2 we had to include the <asm/termios.h> header. We would have to figure out a way to have both termios together, or on linux have all files call <asm/termios.h> instead of <termios.h>

matthova:
Not sure how I got this to work in the first place. I'm currently failing with this error:
```
node: symbol lookup error: /home/pi/serialport/build/Release/serialport.node: undefined symbol: _Z5ioctliiz
```

src/custom_baudrate.cpp
reconbot: if you can move the ifdef into unix.cpp and include one for `__linux__` and another for .. not linux

matthova:
I won't be able to focus on this for another two weeks. Till then, if anyone wants to try tweaking this PR, give it a go 👍

matthova:
I won't be able to focus on this for another two weeks. Till then, if anyone wants to try tweaking this PR, give it a go 👍

D1plo1d:
*bump* Any update on this?

250,000 Baud is the default baud rate used by the [Marlin](https://marlinfw.org) 3D printer firmware and without it my project [Tegh](https://github.com/tegh/tegh) cannot support the latest firmware.

Fumon:
I put some work into this. I have the custom baud rate implementation from here cleaned up and integrated but I found an issue with the tests for custom baud rate setting so there's a lot more in my patches.

#1464

PR #1448, opened 2018-01-21
Adding in port selection to serialport-term.

robertmassaioli:
This makes the terminal command much more usable when you need to open
it multiple times in a row; especially when the names of the serial ports
are rapidly changing.

screen shot 2018-01-21 at 1 30 35 pm

At the moment, I'm doing a bunch of development on the nRF5 series of SOC's from Nordic Semiconductor and, because this is Bluetooth, I often have more than one SOC connected to my computer. Also, every time that I plug a new device in the [Silicon Labs driver](https://www.silabs.com/products/development-tools/software/usb-to-uart-bridge-vcp-drivers) gives me a new virtual port for the device; making it a more lengthy than required process to reconnect to my logs. It would be better if it would just be:

 1. Ctrl-C 
 1. Rerun terminal
 1. Select the newly named port.
 1. Profit

At any rate, hopefully, you like these changes.

### DIFF
--- a/bin/terminal.js
+++ b/bin/terminal.js
@@ -46,7 +46,7 @@ function setupPort() {
       console.error('Error listing ports, and missing port argument.', err);
       args.outputHelp();
       args.missingArgument('port');
-      process.exit(-1);
+      process.exit(4);
     } else {
       if (ports.length > 0) {
         const portSelection = new List({
@@ -63,12 +63,12 @@ function setupPort() {
           })
           .catch(error => {
             console.log(`Could not select a port: ${error}`);
-            process.exit(-2);
+            process.exit(2);
           });
       } else {
         args.outputHelp();
         args.missingArgument('port');
-        process.exit(-1);
+        process.exit(3);
       }
     }
   });
--- a/bin/terminal.js
+++ b/bin/terminal.js
@@ -37,7 +37,7 @@ function listPorts() {
 };
 
 function setupPort() {
-  if(args.port) {
+  if (args.port) {
     createPort(args.port);
   }
 
@@ -48,13 +48,13 @@ function setupPort() {
       args.missingArgument('port');
       process.exit(-1);
     } else {
-      if(ports.length > 0) {
-        var portSelection = new List({
+      if (ports.length > 0) {
+        const portSelection = new List({
           name: 'serial-port-selection',
           message: 'Select a serial port to open',
           choices: ports.map((port, i) => `[${i + 1}]\t${port.comName}\t${port.pnpId || ''}\t${port.manufacturer || ''}`)
         });
-  
+
         portSelection.run()
           .then(answer => {
             const choice = answer.split('\t')[1];
--- a/bin/terminal.js
+++ b/bin/terminal.js
@@ -4,6 +4,7 @@
 const SerialPort = require('../lib/');
 const version = require('../package.json').version;
 const args = require('commander');
+const List = require('prompt-list');
 
 function makeNumber(input) {
   return Number(input);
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -62,6 +62,7 @@
     "nan": "^2.6.2",
     "prebuild-install": "^2.4.1",
     "promirepl": "^1.0.1",
+    "prompt-list": "^3.1.2",
```
robertmassaioli: This change does require a new dependency. I don't know if this will be contentious or not. Please let me know.

package.json
```diff
@@ -99,6 +100,7 @@
     "lint": "eslint lib test bin examples",
     "rebuild-all": "npm rebuild && node-gyp rebuild",
     "repl": "node bin/repl.js",
+    "terminal": "node bin/terminal.js",
```
robertmassaioli: This is useful for development because of: https://github.com/npm/npm/issues/7137

bin/terminal.js
```diff
@@ -4,17 +4,17 @@
...
 const args = require('commander');
+const List = require('prompt-list');
 
 function makeNumber(input) {
   return Number(input);
 }
 
 args
   .version(version)
-  .usage('-p <port> [options]')
```
robertmassaioli: Since port is no longer a required argument we can remove references to it.

bin/terminal.js
```diff
@@ -36,21 +36,53 @@ function listPorts() {
...
+      if(ports.length > 0) {
+        var portSelection = new List({
+          name: 'serial-port-selection',
+          message: 'Select a serial port to open',
+          choices: ports.map((port, i) => `[${i + 1}]\t${port.comName}\t${port.pnpId || ''}\t${port.manufacturer || ''}`)
+        });
+  
+        portSelection.run()
+          .then(answer => {
+            const choice = answer.split('\t')[1];
```
robertmassaioli: This is probably the hackiest part of the whole PR. Relying on the choices to be tab separated in order to extract the comName.

I would prefer it if the prompt-list library had a `name` and `value` fields so that what you display could be different to what gets returned as an 'answer' but, alas, it does not look like that is the case.

bin/terminal.js
```diff
@@ -36,21 +36,53 @@ function listPorts() {
...
+            console.log(`Opening serial port: ${choice}`);
+            createPort(choice);
+          })
+          .catch(error => {
+            console.log(`Could not select a port: ${error}`);
+            process.exit(-2);
```
robertmassaioli: Not sure what your policy is on the error code numbering that should be returned. Could use feedback.

robertmassaioli:
I looked at listr on your recommendation and I think that prompt-list is a better fit for this use case. But maybe my cursory glance did not reveal something in listr that makes it a better fit.

I think the error codes look good based on what you have said. What would you like me to do in order for this PR to be merged in? It looks good to me. Cheers!

bin/terminal.js
reconbot: Lets make it a positive int for fewer surprises cross platform

robertmassaioli:
Awesome, I've modified all of the error codes so that they are positive. Even the one that was `-1` before. 

I'll wait till you get back next week. Cheers, have a great break, and thank you for your time! 😄

robertmassaioli:
Hi @reconbot , I understand that you are on vacation so I don't want to ping too much, but by which date would you be available for us to pick this PR up and see if we can get it over the line?

robertmassaioli:
Thank you! Excited to see these changes go out in the next release! Your rock, keep up the great work!

PR #1464, opened 2018-02-04
WIP: Custom baud rate refactor (linux) and testing (all platforms)

Fumon:
Forked from #1388, cleaned it up a bit and added more testing

The test environment didn't actually have a way to query the system to check the set baud rate so I added that to the bindings and rewrote those tests. Tested on linux with mock and arduino echo. Didn't have the build environment on windows to test but uses existing get interface on win32 so I can't imagine it doesn't work.

I added the GetBaudRate implementation for linux and windows but I have no mac hardware (or a paid dev licence) so I couldn't consult the Mac documentation for IOKit to add that in as well. This patch will break the test on darwin until it's implemented. If it can't be quickly added, I'll submit a new PR separate from the tests. I felt it was important to highlight that this functionality was broken on linux and the current tests didn't catch it.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -39,7 +39,8 @@
         {
           'sources': [
             'src/serialport_unix.cpp',
-            'src/poller.cpp'
+            'src/poller.cpp',
+            'src/custom_baudrate.cpp'
           ]
         }
       ]
--- a/binding.gyp
+++ b/binding.gyp
@@ -35,12 +35,20 @@
           }
         }
       ],
-      ['OS!="win"',
+      ['OS=="linux"',
         {
           'sources': [
             'src/serialport_unix.cpp',
             'src/poller.cpp',
-            'src/custom_baudrate.cpp'
+            'src/serialport_linux.cpp'
+          ]
+        }
+      ],
+      ['OS!="win"',
+        {
+          'sources': [
+            'src/serialport_unix.cpp',
+            'src/poller.cpp'
           ]
         }
       ]
--- a/lib/bindings/base.js
+++ b/lib/bindings/base.js
@@ -188,6 +188,20 @@ The in progress writes must error when the port is closed with an error object t
   }
 
   /**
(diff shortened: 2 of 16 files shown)


Fumon:
All the tests are passing now. Sorry for the commit spam, needed the CI for windows and mac.

I separated the system level check in the testing harness on darwin so this should be good to go. Let me know if there are any style conventions or other issues.

I had the same thought about "GetBaudRate" being kinda clunky next to all the other single word get/set/open/list commands. I'd be willing to undertake rolling it into Get or possibly make a verbose "stat" call but that's, as you said, another PR.

Fumon:
Sounds good.

WRT OS X: I found some more docs tonight while writing this reply. The main issue is that the custom baud rate setting on OSX that we're using (although I haven't found any alternative to using it yet since PySerial uses the same interface [here's the line](https://github.com/pyserial/pyserial/blob/master/serial/serialposix.py#L190)) does not update the termios struct with the new baudrate (cited [here](https://discussions.apple.com/thread/1587991?start=0&tstart=0)) and also does not have any additional ioctl's for retrieving baud rates (based on investigating [this header](https://opensource.apple.com/source/IOSerialFamily/IOSerialFamily-32.2/IOSerialFamily.kmodproj/ioss.h.auto.html) and related files). However, based on the implementation of IOSSIOSPEED [here](https://opensource.apple.com/source/IOSerialFamily/IOSerialFamily-32.2/IOSerialFamily.kmodproj/IOSerialBSDClient.cpp.auto.html) it appears that the situation on Linux (where the previous hack was silently failing) is very unlikely to occur.

So in summary: There doesn't seem to be a way to retrieve the baud rate if it's set to something custom on Darwin, at least through the interface we're using. There could be other interfaces hiding in OS X but I haven't seen them.

matthova:
@Fumon 🙌  awesome. I need to try this out this weeked :) Great PR!

PR #1482, opened 2018-02-14
Fix examples on case sensitive filesystems

danmichaelo:
On case sensitive filesystems (Windows), this did not work:
```
const Readline = SerialPort.parsers.Readline;
```
This did:
```
const Readline = SerialPort.parsers.readline;
```

I'm not too familiar with Node, so I'm not sure what's the role of `index.js` here, but at least here's a suggested fix.


danmichaelo:
My mistake, I was using an old version!

PR #1483, opened 2018-02-15
Query registry for USB serial numbers in win

### LINKED ISSUE #1459: Wrong serial number on CDC ACM composite devices on win32
This is kind of a follow-up of #1220.

I'm using `serialport` v6.0.4 (latest available as of 2018-01-28) on nodejs 8.9. I have a Linux (debian sid) box, and a win32 (Windows 10 v1607 build 14393.2007) box.

The serial port device I'm working with is a [nRF52840 preview development kit](http://infocenter.nordicsemi.com/topic/com.nordic.infocenter.nrf52/dita/nrf52/development/nrf52840_pdk/kit_hw_content.html?cp=2_0_3_2_0), loaded with firmware [implementing USB CDC ACM](http://infocenter.nordicsemi.com/topic/com.nordic.infocenter.sdk5.v14.2.0/lib_usbd_class_cdc.html?cp=4_0_0_3_55_8_3) for serialport-like communications. It's basically a beefed-up arduino-like board with wireless stuff and its own fully-configurable USB stack.

### Summary of Problem

In my case, the serial number of the USB device is set up when compiling the firmware, and is used by the USB stack to populate the `iSerial` field of the USB descriptor. This is how the serial number looks over the wire (wireshark screenshot):

![screenshot_20180129_142425](https://user-images.githubusercontent.com/1125786/35512526-52766024-0500-11e8-9635-5ea4ab4df204.png)

And this is how a `lsusb -v` looks like in my linux box (relevant part):

```
Device Descriptor:                                                                                                                                                                                                                                             
  idVendor           0x1915 Nordic Semiconductor ASA                                                                                                                                                                                                                          
  idProduct          0x521f                                                                                                                                                                                                                                                   
  bcdDevice            1.00                                                                                                                                                                                                                                                   
  iManufacturer           1 Nordic Semiconductor                                                                                                                                                                                                                              
  iProduct                2 nRF52 USB CDC Demo                                                                                                                                                                                                                                
  iSerial                 3 bf8f64cd7b5be46c             
```

And this is how `node-serialport/bin/list.js -f jsonline` looks like in linux (with extra line breaks for reading pleasure):

```json
{
"manufacturer":"Nordic Semiconductor",
"serialNumber":"bf8f64cd7b5be46c",
"pnpId":"usb-Nordic_Semiconductor_nRF52_USB_CDC_Demo_bf8f64cd7b5be46c-if00",
"vendorId":"1915",
"productId":"521f",
"comName":"/dev/ttyACM0"
}
```

So in Linux, I get to see the serial number of the USB device as the serial number. This is good. This is expected.

Now I switch to win32.

The output of  `node-serialport/bin/list.js -f jsonline` in win32 looks like:

```json
{
"comName":"COM5",
"manufacturer":"Microsoft",
"serialNumber":"7&307B223F&1&0000",
"pnpId":"USB\\VID_1915&PID_521F&MI_00\\7&307B223F&1&0000",
"locationId":"0000.0014.0000.005.002.000.000.000.000",
"vendorId":"1915",
"productId":"521F"}
```

Win32 reports the serial number as `7&307B223F&1&0000`, but it should be `bf8f64cd7b5be46c`... so where does that come from?

The answer is in how win32 handles composite USB devices. If I open up win's device manager, it looks like:

![image](https://user-images.githubusercontent.com/1125786/35561329-36ccd74e-05b0-11e8-86ee-730ac5e06092.png)

If I fire up `regedit.exe` to see details of those two devices (under `HKEY_LOCAL_MACHINE/System/CurrentControlSet/Enum/USB`, I can see...

![image](https://user-images.githubusercontent.com/1125786/35561489-bf7c773e-05b0-11e8-81b4-578ba7937c6f.png)

Well, that makes sense. One of the devices maps to the USB device as a whole, and the other maps to the CDC ACM interface. And the naming makes sense according to https://docs.microsoft.com/en-us/windows-hardware/drivers/usbcon/enumeration-of-interfaces-not-grouped-in-collections .

Digging a bit deeper, I can see how the drivers are bound to "devices":

![image](https://user-images.githubusercontent.com/1125786/35564993-2e87b796-05bc-11e8-9d41-804dbf41f132.png)

So my guess is that `serialport` is interacting with the `usbser.sys` kernel driver, fetching information from it - but the right serial number is not available at that level.

The question is: **How can `serialport` check if a serial port is part of a USB composite device, and if so how can it fetch the parent's serial number?**

Get a hold of a USB composite device with a CDC ACM interface, play around with it with the windows device manager as shown above :-)

---

Addendum: It is possible to force win32 to bind the `usbser.sys` to the USB device as a whole: run Zadig, show all devices, choose the composite device, replace `usbccgp.sys` with `usbser.sys` (AKA `USB Serial (CDC)`)". In this case, things will look like:

![image](https://user-images.githubusercontent.com/1125786/35565349-839f52ba-05bd-11e8-9f28-3516c83900d6.png)

...and `serialport` will be able to fetch the right serial number. However, this prevents other interfaces in the device from working (e.g. a device with CDC ACM and mass storage at the same time), and requires manual user intervention with scary messages, so it's just a stopgap measure.

IvanSanchez:
Fixes #1459.

This PR makes the code for win32 fetch the [ContainerID](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/container-ids) of the device node holding the serial port, then make a registry lookup for the composite USB device with a matching ContainerID. The serial number from the device node of the composite device is then returned and used.

The code might not be the cleanest ever, but it's in a working state.

Special kudos to @CoqRogue for all the work put on this.

### DIFF
--- a/lib/bindings/win32.js
+++ b/lib/bindings/win32.js
@@ -9,7 +9,7 @@ class WindowsBinding extends BaseBinding {
     return promisify(binding.list)().then(ports => {
       // Grab the serial number from the pnp id
       ports.forEach(port => {
-        if (port.pnpId) {
+        if (port.pnpId && !port.serialNumber) {
           const serialNumber = serialNumParser(port.pnpId);
           if (serialNumber) {
             port.serialNumber = serialNumber;
--- a/src/serialport_win.cpp
+++ b/src/serialport_win.cpp
@@ -6,10 +6,17 @@
 #include <string.h>
 #include <windows.h>
 #include <Setupapi.h>
+#include <initguid.h>
+#include <devpkey.h>
 #include <devguid.h>
 #pragma comment(lib, "setupapi.lib")
 
+#define ARRAY_SIZE(arr)     (sizeof(arr)/sizeof(arr[0]))
+
 #define MAX_BUFFER_SIZE 1000
+
+// As per https://msdn.microsoft.com/en-us/library/windows/desktop/ms724872(v=vs.85).aspx
+#define MAX_REGISTRY_KEY_SIZE 255
 
 // Declare type of pointer to CancelIoEx function
 typedef BOOL (WINAPI *CancelIoExType)(HANDLE hFile, LPOVERLAPPED lpOverlapped);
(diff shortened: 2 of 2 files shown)


lib/bindings/win32.js
```diff
@@ -9,7 +9,7 @@ class WindowsBinding extends BaseBinding {
     return promisify(binding.list)().then(ports => {
       // Grab the serial number from the pnp id
       ports.forEach(port => {
-        if (port.pnpId) {
+        if (port.pnpId && !port.serialNumber) {
```
reconbot: 👍

src/serialport_win.cpp
```diff
@@ -597,6 +604,160 @@ NAN_METHOD(List) {
...
+    // going on.
+    StringFromGUID2((REFGUID)szWUuidBuffer, containerUuid, ARRAY_SIZE(containerUuid));
+  } else {
+    // Container UUID could not be fetched, return empty serial number.
+    return;
+  }
+  
+  // NOTE: Devices might have a containerUuid like {00000000-0000-0000-FFFF-FFFFFFFFFFFF}
+  // This means they're non-removable, and are not handled (yet).
+  // Maybe they should inherit the s/n from somewhere else.
```
reconbot: What's non-removable?

src/serialport_win.cpp
```diff
@@ -634,8 +796,8 @@ void EIO_List(uv_work_t* req) {
     dwSize = sizeof(szBuffer);
     SetupDiGetDeviceInstanceId(hDevInfo, &deviceInfoData, szBuffer, dwSize, &dwSize);
     szBuffer[dwSize] = '\0';
-    pnpId = strdup(szBuffer);
-
+    pnpId = strdup(szBuffer); 
+
```
reconbot: nit: whitespace

src/serialport_win.cpp
IvanSanchez: That means that the serial port is not part of what win32 considers a removable device. In other words: the RS-232 connector soldered to your motherboard.

IvanSanchez:
@reconbot Just test that this works for whatever USB serial adapters you have lying around, specially if you have two or more of the same brand/model. This works for all the stuff in my test bench, but I'm not sure if there is any edge case that we overlooked.

Luckily @CoqRogue did an over-the-shoulder review of the c++ code, but we'll appreciate extra eyes looking at it.

IvanSanchez:
Oh, the joys of linting. I'll have a look.

IvanSanchez:
Rebased, linted, and squashed.

IvanSanchez:
:tada: 

Given the amount of rebasing and linting involved in getting this merged, I'd rather go into my usual "if it works, don't touch it" stance. :grin:

PR #1486, opened 2018-02-15
Fix test timeouts and change test jig

Fumon:
This patch solves a number of timeout bugs in the binding and serial integration tests on linux (did not check on other platforms) which were preventing full passes.

It also changes the advised test setup from an Arduino with the arduinoEcho.ino sketch to any serial device with the rx and tx pins bridged excluding devices that use the ACM driver on linux.

The old test rig also still works aside from the ACM driver versions of arduino due to the ACM driver hanging on close instead of erroring (expected behavior but there will be another patch addressing this). This particular failing test means that the integration test likely never worked for Arduino UNOs and above that all use ttyACMx for interfacing.

Review and comments welcome.

### DIFF
--- a/test/bindings.js
+++ b/test/bindings.js
@@ -427,7 +427,8 @@ function testBinding(bindingName, Binding, testPort) {
         }
 
         let binding;
-        beforeEach(() => {
+        beforeEach(function () {
+          this.timeout(20000);
           binding = new Binding({
             disconnect
           });
--- a/test/bindings.js
+++ b/test/bindings.js
@@ -557,7 +557,8 @@ function testBinding(bindingName, Binding, testPort) {
         beforeEach(() => {
           buffer = Buffer.alloc(readyData.length);
           binding = new Binding({ disconnect });
-          return binding.open(testPort, defaultOpenOptions);
+          return binding.open(testPort, defaultOpenOptions)
+            .then(() => binding.write(buffer));
         });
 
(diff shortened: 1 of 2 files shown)


Fumon:
Could you be more specific? I found that most of the tests in the integration suite didn't pass for an Arduino Uno with the ArduinoEcho sketch for one reason or another. Timeouts and the ACM driver on Linux both were wreaking havoc. The parts of the code which use the ready parser were not really getting reached in my tests due to these issues and when they were, they didn't pass.

The advantage of using the RXTX short is that it's instant regardless of baudrate so there's no received data lag at all and we can test at any baud rate. The OS will dutifully hand you the data until you close or flush the data which more accurately tests the kind of behaviour we're after in real life.

At least one test (the 2k test) seemed to depend on the slow reaction speed of the arduino echo loop and needed a flush under any other serial situation.

The "errors if closing during a write" test does not complete on an Arduino Uno or above due to the ACM driver not closing while writes are occurring at the kernel level. If you give that test a really long timeout and watch the debug, it successfully closes without error eventually (which fails the test). Not sure what to do here since it's intended behaviour?

The test suite will still run fine with the ArduinoEcho sketch because every time I manually write "READY", the device will trigger the ready parser itself instead of echoing more data. Actually, Arduinos that use ACM still need to use the ArduinoEcho as the baudrate looks like it's more strictly enforced on cdc_acm. Still working that out.

PR #1510, opened 2018-03-02
Avoid using deprecated Buffer constructor

ChALkeR:
`safe-buffer` is already used and provides `Buffer.alloc` polyfill, so just use `Buffer.alloc` directly to avoid hitting deprecated API.

Refs:
https://nodejs.org/api/deprecations.html#deprecations_dep0005_buffer_constructor

Tracking: https://github.com/nodejs/node/issues/19079

### DIFF
--- a/lib/bindings/mock.js
+++ b/lib/bindings/mock.js
@@ -17,7 +17,7 @@ class MockBinding extends BaseBinding {
     this.isOpen = false;
     this.port = null;
     this.lastWrite = null;
-    this.recording = new Buffer(0);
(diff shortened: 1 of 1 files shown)


ChALkeR:
@reconbot Yes, this has been done before in 5f53a45cae0d7b489b3bc3e5eb4557526fa88e7b but slipped in again in e8d9d2b16c664224920ce1c895199b1ce2def48c due to a linter issue. See https://github.com/mysticatea/eslint-plugin-node/pull/110 for context.

PR #1513, opened 2018-03-05
feat(parsers): Add a "Throughput Meter" stream transformer

IvanSanchez:
I found myself needing some kind of throughput meter for one of my `Stream`s, and I thought "well, maybe this is interesting for `node-serialport` too. So here it is.

I'm also wondering if these stream transformers would make better sense in a separate repo / package, since they are useful for other streams as well, not only serial port streams.


IvanSanchez:
I'm partial to creating a monorepo for this kind of transform parsers. Not sure if it's the best way forward, though.

PR #1576, opened 2018-06-11
K4connect/usb description

jonathan-k4:
Added support for USB description field for both Linux and Darwin

### DIFF
--- a/examples/list.js
+++ b/examples/list.js
@@ -1,0 +1,12 @@
+/* eslint-disable node/no-missing-require */
+'use strict';
+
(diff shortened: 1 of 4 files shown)


jonathan-k4:
I think its still helpful to surface ID_MODEL as the description. As it is programmable in the field, it can be used to differentiate between multiple USB devices of the same time. In a control system, we often program the description field of USB serial adapters with an identifying string unique to that cable. So when we multiple on the same control system, surfacing this field lets us know which serial device is which in node.

PR #1602, opened 2018-07-21
put back prebuild binary builders

reconbot:
I'll use my branch that fixes `prebuild --all` until it lands and is released.


monkbroc:
Merci beaucoup!

PR #1693, opened 2018-10-05
feat: use bl for a faster delimiter parser

reconbot:


HipsterBrown:
Any metrics for speed improvements? Is it possible to add some information in the description of this PR as to why `bl` is faster?

HipsterBrown:
Cool, just want to add some context to the PR for people curious about the addition of this dependency. 😄

PR #1699, opened 2018-10-12
Fixed uninitialized outbaud

PhilLab:
data->baudRate was assigned an unassigned outbaud value.

Another (maybe better) solution would be to give outbaud a default value. should that be 0 or -1 ? Creating this pull request as basis for the discussion


PhilLab:
I am referring to this piece of code: https://github.com/node-serialport/node-serialport/blob/2e93c5556483d56c5cf589d0cbd2da4fd3132ef2/packages/bindings/src/serialport_unix.cpp#L369-L381

``int outbaud; `` was not always initialized. So maybe ``int outbaud = -1;`` would solve the issue with less intrusive changes?

PhilLab:
Well, my first proposal was to handle it like the missing OSX implementation is already handled: ``snprintf`` and then ``return``. I don't have a strong opinion on that, you as maintainer can choose :-)

PR #1707, opened 2018-10-23
add orient xwt-ps050 serialports name prefix

digdas:
Hello!
I work with device orient xwt-ps050 in centos7, and it have "WCH" serialports name prefix.
I found no other way to add it without lost in updates, if it possible to use some public api to change this checkPathOfDevice method tell me.
Thank you!


digdas:
Thanks!

PR #1737, opened 2018-12-05
doc: update layout and homepage

reconbot:
This adds some usage examples and general info of why you'd use serialport to the homepage.

I also combined the guides and docs sections so it's easier to find information.


packages/website/static/css/custom.css
```diff
@@ -28,3 +28,11 @@
...
   text-decoration: underline;
 }
+
+.imageAlignTop .blockImage {
+  max-width: 300px;
+}
+
+code {
+  text-align: left;
+}
```
HipsterBrown:
```suggestion
}

.gridBlock {
  align-items: flex-end;
}

.twoByGridBlock {
  margin-bottom: 36px;
}
```

This helps align the content of the grid blocks. But not a blocker for shipping. 🚀
screen shot 2018-12-05 at 11 01 58 am

PR #1740, opened 2018-12-09
Fix codeblock syntax highlighting in docs

dsanders11:
Wasn't marked as `js` like the other blocks, making the syntax highlighting look off and highlighting words in the comment.


dsanders11:
@reconbot, this should be a quick merge if you're not too busy.

PR #1741, opened 2018-12-09
Fix syntax highlighting in docs

dsanders11:
The code in this block wasn't being highlighted, and in other parts of the doc `typescript` is specified instead of `ts`, so assuming `ts` is a typo stopping it from highlighting properly.

### DIFF
--- a/docs/api-binding-abstract.md
+++ b/docs/api-binding-abstract.md
@@ -5,7 +5,7 @@ title: Abstract Binding
 
 Typescript signature of Abstract Binding
 
-```ts
+```typescript
 
 class AbstractBinding {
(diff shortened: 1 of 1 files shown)


dsanders11:
@reconbot, hmm, strange. If I edit the page [here](https://github.com/node-serialport/node-serialport/edit/master/docs/api-binding-abstract.md) it actually highlights correctly under preview as is, and still does if I change it to `typescript`. Maybe there's some other issue preventing it from highlighting on the hosted docs? Not sure, I assumed it was changing it to `typescript` that made the preview render correctly for me, but it looks like `ts` does work as an alias. Although, for consistency it'd probably be best to use `typescript` like other parts of the docs.

PR #1748, opened 2018-12-20
PUBLISHING: update with details about prebuilding binaries

HipsterBrown:


PUBLISHING.md
```diff
@@ -10,8 +10,26 @@ Two factor auth is required for publishing.
...
 
-For example for version 2.0.3
+CI currently expects the latest git tag to match the branch it is building before kicking off the prebuild process.
+
+For example for version 2.0.3:
+
+Check the latest tag.
+
+```
+git describe --tag --always HEAD
```
reconbot: this only returns the first tag on the commit, not all of them

HipsterBrown: So this is what CI uses currently. Should we change it to use `git tag --points-at HEAD` instead so we don't have this issue of repushing the `@serialport@bindings` tag?

reconbot: good catch, I understand this now a lot better

HipsterBrown:
The new logic in appVeyor.yml needs to be verified on Windows before merging.

PUBLISHING.md
```diff
@@ -10,8 +10,25 @@ Two factor auth is required for publishing.
 
 If publishing more than 3 packages at once and one of them is the bindings package, you'll need to figure out a way to get the ci's to build the binaries as github wont tell them about the new tags. You can do this by deleting the tag and pushing it again for the bindings package (only binary package as of this writing).
 
-For example for version 2.0.3
+CI currently expects the latest git tag to match the branch it is building before kicking off the prebuild process.
```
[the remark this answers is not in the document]
HipsterBrown: Good catch. I'll update that.

PR #1765, opened 2019-01-04
fix: propagate async context in callbacks

### LINKED ISSUE #1751: receiving data delayed / loop issues
## prelude

i have a problem which is a bit hard to explain. but i'll try:

i have a simple express+socket.io server which acts as a serialport to websocket gateway.
it is part of a lib i'm developing: https://github.com/enocean-js/enocean-js

the server code can be found here: https://github.com/enocean-js/enocean-js/blob/master/packages/node_modules/%40enocean-js/websocket-gateway/server.js

if you clone the repo, and `npm i`, everything works as expected...
but when i remove my package.lock and then `npm i` the Bug show up.

i'm using my own parser: https://github.com/enocean-js/enocean-js/blob/master/packages/node_modules/%40enocean-js/serialport-parser/src/serialport-parser.js

## Bug

when receiving data on the serialport, the parser.on("data",...) event does not fire when the parser finished parsing the input, but after a few seconds, all detected packets/events are emitted. 

in the meantime i can still write to the port without a problem.

now here comes the part i hope will help you find the problem: when i insert a timer

    setInterval(()=>{},0)

everthing works again as expected... moreover, if i set the interval to any  value, the events are fired at exactly that interval...

i hope this gives you a hint at where the problem might be...

## p.s.

i'm on linux (Arch)
tried with different node versions (my min-version is 8.4.0) up to 11.4.0.

zbjornson:
Fixes #1751 

I've tested this on Windows with real hardware (repro'ed the issue in #1751). I can't get [WSL](https://docs.microsoft.com/en-us/windows/wsl/faq) to work with a serial port right now to test on Ubuntu.

I can't think of a good way to make this commit smaller. `AsyncResource` needs to be constructed with a `const char*` name unique to each of your `struct`s (so can't inherit from some other class that extends `AsyncResource`). Constructing the `AsyncResource` (in a ctor or initializer list) makes it necessary to initialize everything else that you were relying on value-initialization for. Rather than only initializing the members that you're not setting where the structs are used, I went with the safer option of initializing (almost) everything explicitly. Using more C++ features (`std::string`s) would shorten the code overall (possibly making initialization unnecessary) but would make this commit even bigger. /ramble

### DIFF
--- a/packages/bindings/src/serialport_win.h
+++ b/packages/bindings/src/serialport_win.h
@@ -9,17 +9,18 @@
 
 #define ERROR_STRING_SIZE 1024
 
-struct WriteBaton {
-  int fd;
-  char* bufferData;
-  size_t bufferLength;
-  size_t offset;
-  size_t bytesWritten;
-  void* hThread;
-  bool complete;
+struct WriteBaton : public Nan::AsyncResource {
+  WriteBaton() : AsyncResource("node-serialport:WriteBaton"), bufferData(), errorString() {}
+  int fd = 0;
+  char* bufferData = nullptr;
+  size_t bufferLength = 0;
+  size_t offset = 0;
+  size_t bytesWritten = 0;
+  void* hThread = nullptr;
+  bool complete = false;
   Nan::Persistent<v8::Object> buffer;
   Nan::Callback callback;
-  int result;
+  int result = 0;
   char errorString[ERROR_STRING_SIZE];
 };
 
@@ -29,15 +30,16 @@ void EIO_AfterWrite(uv_async_t* req);
 DWORD __stdcall WriteThread(LPVOID param);
 
 
-struct ReadBaton {
-  int fd;
-  char* bufferData;
-  size_t bufferLength;
-  size_t bytesRead;
-  size_t bytesToRead;
-  size_t offset;
-  void* hThread;
-  bool complete;
+struct ReadBaton : public Nan::AsyncResource {
+  ReadBaton() : AsyncResource("node-serialport:ReadBaton"), errorString() {}
+  int fd = 0;
+  char* bufferData = nullptr;
+  size_t bufferLength = 0;
+  size_t bytesRead = 0;
+  size_t bytesToRead = 0;
+  size_t offset = 0;
+  void* hThread = nullptr;
+  bool complete = false;
   char errorString[ERROR_STRING_SIZE];
   Nan::Callback callback;
 };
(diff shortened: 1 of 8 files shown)


zbjornson:
Urgh, those CI failures are due to [a GCC bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=43453) that's fixed in 4.10.0, I think (see https://travis-ci.org/node-serialport/node-serialport/builds/475554563, which used GCC5 and was successful).

I can't figure out how [NAN's AsyncResource test](https://github.com/nodejs/nan/blob/master/test/cpp/asyncresource.cpp), which uses the same syntax, builds with 4.8.5 (e.g. https://travis-ci.com/nodejs/nan/jobs/166028162). Anyone see something I'm missing?

fivdi:
It looks like the compile error messages are pointing at the wrong lines of code and that it's actually the non-static data member initializers for strings [like this one](https://github.com/node-serialport/node-serialport/blob/8841debc090d0ed4e457912e8e32612f87671b1d/packages/bindings/src/serialport.h#L62) or [this one](https://github.com/node-serialport/node-serialport/blob/8841debc090d0ed4e457912e8e32612f87671b1d/packages/bindings/src/serialport.h#L64) that are causing the errors.

There's an online compiler for doing some tests with gcc 4.8, for example, [this doesn't work](http://coliru.stacked-crooked.com/a/e69073ac4196ed2c) and shows the compile error `"array used as initializer"`, but [this works](http://coliru.stacked-crooked.com/a/c71a81e7ce699eb3) and displays `"all ok"`.

If the non-static data member initializers for strings are replaced with code in the constructor I think it will work.

packages/bindings/src/serialport_win.h
```diff
@@ -62,10 +64,11 @@ struct ListResultItem {
   std::string productId;
 };
 
-struct ListBaton {
+struct ListBaton : public Nan::AsyncResource {
+  ListBaton() : AsyncResource("node-serialport:ListBaton") {}
   Nan::Callback callback;
   std::list<ListResultItem*> results;
-  char errorString[ERROR_STRING_SIZE];
+  char errorString[ERROR_STRING_SIZE] = "";
```
reconbot: Why is this one ok when the others weren't? Because it's for windows?

zbjornson: Yeah. It's valid c++11 and the oldest version of MSVC that Node.js builds with is fine with it. If you want them to be consistent I'll happily change it.

reconbot: I am uncertain what a uninitialized char array looks like, so I don't know if this causes any different behaviors between the two different ways in how we use it. If it's just aesthetic then sure let's make it consistent.

zbjornson: They're equivalent. Just pushed so they're consistent.

PR #1774, opened 2019-01-10
Change signature of setBaudRate to avoid creating new baton

### LINKED ISSUE #1771: Using SerialPort ends in Segmentation fault (Speicherzugriffsfehler)

zbjornson:
This is an alternative to #1772 that fixes #1771. I don't have a linux box handy to test this with, fyi.

In a moment I'll open another PR that uses `std::string` instead for comparison.

### DIFF
--- a/packages/bindings/src/serialport.h
+++ b/packages/bindings/src/serialport.h
@@ -131,5 +131,4 @@ struct VoidBaton : public Nan::AsyncResource {
 };
 
 int setup(int fd, OpenBaton *data);
-int setBaudRate(ConnectionOptionsBaton *data);
 #endif  // PACKAGES_SERIALPORT_SRC_SERIALPORT_H_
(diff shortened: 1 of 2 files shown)


zbjornson:
> In a moment I'll open another PR that uses std::string instead for comparison.

Ugh, every approach I start on ends up being a bigger change than I'm comfortable making, since I don't have a way to test all of this library and CI didn't catch the original bug. I also made a mistake in this PR and am out of time right now to fix it. Can come back to it, but otherwise #1772 works.

PR #1804, opened 2019-02-14
Fix: Don't restart polling when libuv emits an error

### LINKED ISSUE #1803: Should the Poller start polling the file handle again if it encounters an error?
### Summary of Problem

I'm trying to connect to a USB device on macOS. I can do so successfully and read data. However, when I manually disconnect the USB cable, the process _sometimes_ gets stuck: CPU usage jumps to 100% and I have to forcefully kill it with a task manager.

The issue is quite rare (happens less than 5% of the time) but is very disruptive as I'm not using an environment where a force-restart of the node process is viable.

I stepped through the bindings C++ code and found the issue here: https://github.com/node-serialport/node-serialport/blob/a80d0b73a1d8424901ab73b314de344dad798d12/packages/bindings/src/poller.cpp#L52-L70

Specifically, I keep seeing `status == -9`, which is libuv's `UV_EBADF` "bad file descriptor" error. Line 67 calls the poller again regardless of whether there was a valid event or an error code. `onData` is called _again_ with the exact same error status very shortly afterwards.

### Code to Reproduce the Issue

```js
const serialport = new SerialPort(path)
const parser = serialport.pipe(new SerialPort.parsers.Readline)
parser.on('data', (data) => console.log(data))

// manually disconnect the USB cable
```

### Versions, Operating System and Hardware

- SerialPort@7.1.4
- Node.js v10.15.1
- Mac
- MacBook Pro (Retina, Mid 2012) 2.6 GHz Intel Core i7

nfrasser:
Fixes #1803 

This prevents an infinite loop of polling on a bad file handle

### DIFF
--- a/packages/bindings/src/poller.cpp
+++ b/packages/bindings/src/poller.cpp
@@ -61,10 +61,10 @@ void Poller::onData(uv_poll_t* handle, int status, int events) {
     // fprintf(stdout, "OnData status=%d events=%d\n", status, events);
     argv[0] = Nan::Null();
     argv[1] = Nan::New<v8::Integer>(events);
+    // remove triggered events from the poll
+    int newEvents = obj->events & ~events;
+    obj->poll(newEvents);
   }
-  // remove triggered events from the poll
-  int newEvents = obj->events & ~events;
-  obj->poll(newEvents);
 
   obj->callback.Call(2, argv, obj);
 }


nfrasser:
I am beta testing this fix with my fork and I'm finding that for some cases, the `'open'` event never fires. The issue is specifically on macOS systems. Still working on getting the debug logs for those cases.

@reconbot any ideas as to why this might happen? It looks like the only option I have left it so force-restart the node process when we detect that serialport gets stuck.

packages/bindings/src/poller.cpp
```diff
@@ -61,10 +61,10 @@ void Poller::onData(uv_poll_t* handle, int status, int events) {
     // fprintf(stdout, "OnData status=%d events=%d\n", status, events);
     argv[0] = Nan::Null();
     argv[1] = Nan::New<v8::Integer>(events);
+    // remove triggered events from the poll
+    int newEvents = obj->events & ~events;
+    obj->poll(newEvents);
```
reconbot: ok yes this wont update the poll events on an error - but no this wont stop the poller on an error we should try to call `uv_poll_stop(poll_handle);`

PR #1809, opened 2019-02-25
Fixed RTS/CTS flow control for Windows

krutkay:
The previous implementation of RTS/CTS flow control would hold the RTS line high and ignore the CTS line.

This corrects those two issues.  When the RTS/CTS option is enabled, the serial port will now not send data to the remote device until the CTS line is low, and will lower the RTS line when the input buffer is more than three-quarters full (as per [this document](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/ns-winbase-_dcb)).

### DIFF
--- a/packages/bindings/src/serialport_win.cpp
+++ b/packages/bindings/src/serialport_win.cpp
@@ -125,7 +125,8 @@ void EIO_Open(uv_work_t* req) {
   }
 
   if (data->rtscts) {
-    dcb.fRtsControl = RTS_CONTROL_ENABLE;
+    dcb.fRtsControl = RTS_CONTROL_HANDSHAKE;
+    dcb.fOutxCtsFlow = TRUE;
   } else {
     dcb.fRtsControl = RTS_CONTROL_DISABLE;
   }


krutkay:
CI is failing because of a [Highlight.js issue](https://github.com/highlightjs/highlight.js/issues/1984), which is a subdependency of the website package.

krutkay:
Sounds good.  I appreciate the effort that you put in to maintaining this library, and understand that you'd be concerned about breaking it (and other people's code).

Let me know if there's anything I can do to help with your testing!

krutkay:
My understanding of flow control (mostly learned from [TLDP](https://www.tldp.org/HOWTO/Text-Terminal-HOWTO-11.html)) is that each port (device) has a "request-to-send" (RTS) output pin, and a "clear-to-send" (CTS) input pin.  The RTS of one device is connected to the CTS of the other device.

The device receiving data will raise its RTS line when the connection is opened (the buffer is empty), indicating that it can receive additional data.  The recipient device will hold the line high until its buffers are (almost) full, at which point it will lower the line; the line stays low until the buffers have (partially) drained, when it will raise the RTS line again to indicate it can receive additional data.

The device that's sending data will monitor the CTS line: if the line is high, it is "cleared" to send data.  When the CTS line goes low, it should not send more data to the receiving device (in reality, it's okay for the sender to transmit a couple additional bytes of data).

If my understanding is incorrect, then please stop reading now, because I'm about to make a fool of myself...

---------

The current (Windows) implementation is essentially equal to disabling flow control and manually setting the RTS pin high.  It will never actually change the flow of data.

- `RTS_CONTROL_ENABLE` permanently sets the RTS line to be high, so we will never tell the remote device to stop sending us data.
- The `fOutxCtsFlow` flag is set to false, so the CTS line is not monitored and we will never stop sending data when the remote buffers are full.

(On a side note, if we use `RTS_CONTROL_DISABLE`, we _should_ never receive any data - I don't think the `ENABLE` and `DISABLE` options are useful flags to have in the library).

I believe the flow control option should be `RTS_CONTROL_TOGGLE` as it will lower the RTS line when the Windows input buffer is 75% full, and will hold the line low until the buffer is at least 50% empty. 
 Changing to `RTS_TOGGLE_CONTROL` (and enabling `fOutxCtsFlow`) changes the library behaviour, but shouldn't adversely affect user experience (assuming their hardware correctly implements flow control):

- If they're sending or receiving small amounts of data, they will never trigger conditions where flow control is necessary, so they will never be affected by these changes.  The new behaviour will mirror the old behavior.

- If they're sending or receiving large amounts of data quickly, they may be affected.  These users may be experiencing data loss due to buffer overruns, whether or not they've noticed.  These changes should fix any data loss issues they've been having, as we will now tell the remote device to stop sending data (or will stop sending data to the remote device).  

TL;DR: I _believe_ this PR changes the Windows behaviour to be correct (and match the Linux and Darwin behaviour).

thedayofcondor:
I tested @krutkay patch on my windows version of serialport@7.1.5 against my code and I can confirm it works as intended - I now get consistent behaviour across Linux and Windows (AKA HW handshaking is working as expected)

thedayofcondor:
I am still testing - but with that patch I managed to reliably establish a connection, download the device configuration for the first time (it's a long multipart back and forth between my pc and the device) and can stream data without losses. All working 100% of the times.

It is a massive improvement.

Just, something is not 100% right yet - communication takes about 20 seconds from Windows, 200ms from Linux, like I was working with a really tiny buffer and the HW handshaking was stopping the communication all the times.

However it may be a completely unrelated issue.

thedayofcondor:
The latest news is... it works. It is as fast as Linux today - I am not sure what the issue was, looks like a reboot fixed it 🙄.

PR #1850, opened 2019-04-28
WIP feat: Async-Iterator interface

### LINKED ISSUE #991: Explore Promises in addition to callbacks
Usurps #261 

We'll be dropping older nodes around version 6, we currently polyfill promises for version 4, so we got them to use.

We'd need tests along with any PR

reconbot:
This is the next generation of serial port interfaces. Streams are never going away but the async iterator interface is much more flexible. It can be combined with async-iterator tools (such as [streaming-iterables](https://www.npmjs.com/package/streaming-iterables)) to make buffers and parsers, and can even be combined with our existing stream based parsers.

This is very experimental. I’ve tried to bring a lot of these changes in https://github.com/node-serialport/node-serialport/tree/reconbot/typescript2 but I haven’t had time for a full typescript rewrite. So maybe this smaller api change lets us get some of these advantages without having to rewrite everything.

## Todo
- [ ] api feedback
- [ ] build chain to support modern javascript
- [ ] docs for website
- [ ] abstract away get/set/update borrowing from https://github.com/node-serialport/node-serialport/issues/1679 for local and remote state and parity with web serial
- [ ] tests

## Example Usage
```js
const { open } = require('@serialport/async-iterator')
const Binding = require('@serialport/bindings')
const ports = await Binding.list()
const arduinoPort = ports.find(info => (info.manufacture || '').includes('Arduino'))
const port = await open(arduinoPort)

// read bytes until close
for await (const bytes of port) {
  console.log(`read ${bytes.length} bytes`)
}

// read 12 bytes
const { value, end } = await port.next(12)
console.log(`read ${value.length} bytes / port closed: ${end}`)

// write a buffer
await port.write(Buffer.from('hello!'))
```

closes #991

### DIFF
--- a/greenkeeper.json
+++ b/greenkeeper.json
@@ -3,6 +3,7 @@
     "default": {
       "packages": [
         "package.json",
+        "packages/async-iterator/package.json",
         "packages/binding-abstract/package.json",
         "packages/binding-mock/package.json",
         "packages/bindings/package.json",
--- a/packages/async-iterator/.npmignore
+++ b/packages/async-iterator/.npmignore
@@ -1,0 +1,4 @@
+.DS_Store
+*.test.js
+*.node-10-test.js
+CHANGELOG.md
--- a/packages/async-iterator/README.md
+++ b/packages/async-iterator/README.md
@@ -1,0 +1,13 @@
+# @serialport/AsyncIterator
+
+This is a node SerialPort project! This package does some neat stuff.
+
+- [Guides and API Docs](https://serialport.io/)
+
+This is why you'd use it.
+
+This is how you use it.
+```js
+const asyncIterator = new AsyncIterator()
+
+```
--- a/packages/async-iterator/lib/index.js
+++ b/packages/async-iterator/lib/index.js
@@ -1,0 +1,88 @@
+const debug = require('debug')('serialport/async-iterator')
+
+/**
+ * An AsyncIterator that does something pretty cool.
+ * @param {Object} options open options
+ * @example ```
+// To use the `AsyncIterator` interface:
+const { open } = require('@serialport/async-iterator')
+const Binding = require('@serialport/bindings')
+const ports = await Binding.list()
+const arduinoPort = ports.find(info => (info.manufacture || '').includes('Arduino'))
+const port = await open(arduinoPort)
+
+// read bytes until close
+for await (const bytes of port) {
+  console.log(`read ${bytes.length} bytes`)
+}
+
+// read 12 bytes
+const { value, end } = await port.next(12)
+console.log(`read ${value.length} bytes / port closed: ${end}`)
+
+// write a buffer
+await port.write(Buffer.from('hello!'))
+```
+*/
+
+/**
+ * Wrap an async function so that subsequent calls are queued behind the previous promise resolving
+ */
+const promiseQueue = func => {
+  const queuedFunc = (...args) => {
+    queuedFunc.previousCall = queuedFunc.previousCall.then(() => func(...args))
+    return queuedFunc.previousCall
+  }
+  queuedFunc.previousCall = Promise.resolve()
+  return queuedFunc
+}
+
+const open = async ({ Binding, readSize = 1024, path, bindingOptions = {}, ...openOptions }) => {
+  const binding = new Binding(bindingOptions)
+  debug('opening with', { path, openOptions })
+  await binding.open(path, openOptions)
+
+  const next = async (bytesToRead = readSize) => {
+    if (!binding.isOpen) {
+      debug('next: port is closed')
+      return { value: undefined, done: true }
+    }
+
+    const readBuffer = Buffer.allocUnsafe(bytesToRead)
+    try {
+      debug(`next: read starting`)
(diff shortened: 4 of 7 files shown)


packages/async-iterator/lib/async-iterator.js
```diff
@@ -0,0 +1,68 @@
+const DefaultBindings = require('@serialport/bindings')
+const debug = require('debug')('serialport/async-iterator')
+/**
+ * A transform stream that does something pretty cool.
+ * @param {Object} options open options
+ * @example ```
+// To use the `AsyncIterator` interface:
+const { open, list } = require('@serialport/async-iterator')
+const ports = await list()
+const arduinoPort = ports.find(info => (info.manufacture || '').includes('Arduino'))
```
ThisIsMissEm: A suggestion: have a `select` method, such as the following:

``` 
const [arduinoPort] = select({ info } => (
  (info.manufacturer || '').includes('Arduino')
), {
  limit: 1
})
```

packages/async-iterator/lib/async-iterator.js
```diff
@@ -0,0 +1,68 @@
...
+const port = await open(arduinoPort)
+
+// read bytes until close
+for await (const bytes of port) {
+  console.log(`read ${bytes.length} bytes`)
+}
+
+// read 12 bytes
+const { value, end } = await port.next(12)
```
ThisIsMissEm: Go with `port.read(byteCount)` as it'll make the interface more familiar for those who know WHATWG Streams

packages/async-iterator/lib/async-iterator.js
reconbot: I know there’s a similar api proposed for web serial/web usb, I’ll go take a look!

packages/async-iterator/lib/async-iterator.js
reconbot: I think a stream's interface would be a good idea

ThisIsMissEm: Oh, it looks like I may be wrong about read(n) vs next()

ThisIsMissEm: So yeah, it should be next() but I don't think it takes any arguments: 

https://jakearchibald.com/2017/async-iterators-and-generators/

reconbot: the spec doesn't want any, but it still works if you have an optional argument

marciot:
So I am porting a Python program that relies very extensively on writing commands via the serial port and getting replies back synchronously. It is not the kind of thing that can be rewritten as callbacks easily. I came across this thread when looking for a better solution.

The async iterator is not exactly what I was hoping for,, so I thought I would share what does seem to be working for me:

```
import SerialPort from 'serialport';

class SequentialSerial {
    // Returns a promise that resolves once the serial port is open and ready.
    // The timeout value specifies how long future calls to read() will wait
    // for data before throwing an exception
    open(port, baud, timeout = 1000) {
        this.dec     = new TextDecoder();
        this.timeout = timeout;
        return new Promise((resolve, reject) => {
            this.serial = new SerialPort(port, {baudRate: baud});
            this.serial.on("open",  err => {if(err) reject(err); else resolve();});
        });
    }

    // Returns a promise that resolves once all output data has been written
    flush() {
        return new Promise((resolve, reject) => {
            this.serial.drain(resolve);
        });
    }

    // Returns a promise that resolves after some data has been written
    write(data) {
        return new Promise((resolve, reject) => {
            const okayToProceed = this.serial.write(data);
            if(okayToProceed) {
                resolve();
            } else {
                this.serial.drain(resolve);
            }
        });
    }
    
    // Returns a promise which resolves when "len" bytes have been read.
    // If encoding is "utf8", a string will be returned.
    read(len, encoding) {
        this.readStart = Date.now();
        const tryIt = (serial, resolve, reject) => {
            var result = this.serial.read(len);
            if(result) {
                if(encoding == 'utf8') {
                    result = this.dec.decode(result);
                }
                resolve(result);
            } else {
                if(Date.now() - this.readStart > this.timeout) {
                    reject("SerialTimeout");
                } else {
                    // Using setTimeout here is crucial for allowing the I/O buffer to refill
                    setTimeout(() => tryIt(this.serial, resolve, reject), 0);
                }
            }
        }
        return new Promise((resolve, reject) => {
            tryIt(this.serial, resolve, reject);
        });
    }
    
    // Returns a promise that resolves after a certain number of miliseconds.
    wait(ms) {
        return new Promise((resolve, reject) => {setTimeout(resolve,ms);});
    }
}

async function doSerial() {
    const serial = new SequentialSerial();
    await serial.open("COM6", 9600);
    console.log("Serial port open");
    await serial.wait(5000);
    console.log("Ready");
    await serial.write("hello\r\n");
    await serial.flush();
    const str = await serial.read(7, 'utf8');
    console.log("Read: ", str);
    console.log("Done");
}

doSerial().catch(err => console.log("Serial error:", err));
```

The whole trick here is to make all the I/O operations into Promises and use ECMA 6 await routines to enforce sequencing. Seems to work well and behaves exactly as I would expect it to work in a sequential language. This way of interacting with a serial port seems natural to me, but I am not a node.js programmer, so that probably plays a part.

PR #1851, opened 2019-04-30
chore: remove node6 support and upgrade codebase

reconbot:
- use async functions
- use Object.entries/Object.values
- use Object spreading
- also remove node9 from the test suite, it's not LTS

BREAKING: bindings now use async functions so they’ll never throw, only reject


packages/serialport/test-manual/many-writes.js
```diff
@@ -63,10 +63,10 @@ function writeOneCommandAtATime(port) {
   });
 }
 
-function writeAndDrain(port) {
-  const command = commands.pop();
+async function writeAndDrain(port) {
+  const command = commands.pop()
   if (!command) {
-    return Promise.resolve(port);
+    port
```
icebob: Here is not missing a `return`?

reconbot: you're probably right, pr?

PR #1871, opened 2019-05-17
feat: linux uses by-id path instead of normal /dev path for .list()

### LINKED ISSUE #1476: List symlinks or by ID-path
- Debian Jessie on Raspberry PI2
Hi,

since with every reboot of raspbian (belive it is the same with other UNIX systems) the USB port assignments are jepardized if several serial devices are attached. Like /dev/ttyUSB0 can be /dev/ttyUSB1
at next boot. It would be good to have the possilbity to list devices by device id path for example:
/dev/serial/by-id/usb-FTDI_FT232R_USB_UART_A50285BI-if00-port0
or also list symlinks. Then a static link can be created via /etc/udev/rules.d

This is what I have:

> pi@rpi:~ $ ls -l /dev/ttyUSB1
> crw-rw---- 1 root dialout 188, 1 Feb 6 10:47 /dev/ttyUSB1
> pi@rpi:~ $ ls -l /dev/ttyRflink
> lrwxrwxrwx 1 root root 7 Feb 6 10:44 /dev/ttyRflink -> ttyUSB1

But /dev/ttyRflink does not get listed by serialport.js

reconbot:
The by-id path is consistent regardless of when the device is plugged in. The normal /dev path can change per device.

closes #1476


HipsterBrown:
@reconbot Does this just need manual testing now? Or are you waiting for a major release to merge?

PR #1887, opened 2019-06-05
Added packet timeout for cctalk parser

### LINKED ISSUE #1886: Cctalk message timeout
### What feature you'd like to see
The time between accepted bytes in single cctalk packet cannot exceed 50ms. I want the internal buffer CCTalkParser.array to be reset if I get bytes after delay of more then 50ms.

## Motivation
I have cctalk devices in my project. CCTalkParser is used to fetch data packets from serial device. It fetches packet LENGTH out of the header bytes, waits until LENGTH bytes is received, and gives back the whole packet.

And sometimes I get corrupted data on my data line. That is ok, I can check if the packet is correct using packet checksum. If the checksum is invalid, I ignore the packet.

The problems begin when I get corrupted LENGTH byte. The borders of different packets are mixed. And there is no way to fix it, I have to reopen the serial device connection to restore communication with my cctalk devices.

## Pitch

I want it to be implemented directly as it is written in [CCTalk protocol specification](https://cctalktutorial.files.wordpress.com/2017/10/cctalkpart1v4-7.pdf) section 11.1.

maxvgi:
Should fix #1886

### DIFF
--- a/packages/parser-cctalk/cctalk.js
+++ b/packages/parser-cctalk/cctalk.js
@@ -21,12 +21,12 @@ class CCTalkParser extends Transform {
   }
   _transform(buffer, _, cb) {
     if (this.maxDelayBetweenBytesMs > 0) {
-      const now = +new Date
+      const now = +new Date()
       if (now - this.lastByteFetchTime > this.maxDelayBetweenBytesMs) {
         this.array = []
         this.cursor = 0
       }
-      this.lastByteFetchTime = now;
+      this.lastByteFetchTime = now
     }
 
     this.cursor += buffer.length
--- a/packages/parser-cctalk/cctalk.js
+++ b/packages/parser-cctalk/cctalk.js
@@ -21,12 +21,12 @@ class CCTalkParser extends Transform {
   }
   _transform(buffer, _, cb) {
     if (this.maxDelayBetweenBytesMs > 0) {
-      const now = + new Date
+      const now = +new Date
       if (now - this.lastByteFetchTime > this.maxDelayBetweenBytesMs) {
         this.array = []
         this.cursor = 0
       }
-      this.lastByteFetchTime = + new Date;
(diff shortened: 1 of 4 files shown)


packages/parser-cctalk/cctalk.js
```diff
@@ -12,12 +12,23 @@ const parser = port.pipe(new CCtalk())
...
+  constructor(maxDelayBetweenBytesMs = 0) {
     super()
     this.array = []
     this.cursor = 0
+    this.lastByteFetchTime = 0
+    this.maxDelayBetweenBytesMs = maxDelayBetweenBytesMs
   }
   _transform(buffer, _, cb) {
+    if (this.maxDelayBetweenBytesMs > 0) {
+      const now = +new Date()
```
reconbot: `Date.now()` is a little more straightforward

packages/parser-cctalk/cctalk.js
```diff
@@ -12,12 +12,23 @@ const parser = port.pipe(new CCtalk())
 parser.on('data', console.log)
  */
 class CCTalkParser extends Transform {
-  constructor() {
+  constructor(maxDelayBetweenBytesMs = 0) {
```
reconbot: `packetTimeout`?

maxvgi: Yes, 50ms is the default. But it may break someone's code. I have set 0 for backward compatibility. Is it better to set it to 50 by default?

reconbot: We can do a major version release so it won’t break anyone’s code without them purposely upgrading.

maxvgi:
Hi, @reconbot !
I have added tests as you asked. And changed the major version of cctalk parser package.

maxvgi:
Please, tell me if I should do something more

PR #1899, opened 2019-07-08
chore: reset info.serialNumber when resetting mock ports

boneskull:
BREAKING CHANGE

Resetting the value in the phony port's `info.serialNumber` property
when `BindingMock.reset()` is called will make its behavior "more
deterministic." This is useful when doing deep-equality checks against a
`BindingMock` instance (or more specifically, a `SerialPort` instance
containing a `BindingMock` instance); judicious use of
`BindingMock.reset()` will mean a developer can expect the value of
`info.serialNumber` to be deterministic when tests are run in any order
or in isolation.

People have a tendency to code their way around stuff like this, so this
is _potentially_ a breaking change for _somebody somewhere._

### DIFF
--- a/packages/binding-mock/binding-mock.js
+++ b/packages/binding-mock/binding-mock.js
@@ -26,6 +26,7 @@ class MockBinding extends AbstractBinding {
   // Reset mocks
   static reset() {
     ports = {}
+    serialNumber = 0
   }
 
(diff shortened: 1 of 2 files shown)


boneskull:
If you hate this, you can still keep the test coverage. :wink:

PR #1912, opened 2019-08-14
Make it possible to compile on vanilla Android

podarok:
See https://github.com/Koenkk/zigbee2mqtt/issues/1861
Adding support for Android and default -fPIC not -fPIE make serial port working on Android + Termux ( https://play.google.com/store/apps/details?id=com.termux )


podarok:
Thank you!!!
Now no need to patch the code to get it working

PR #1926, opened 2019-08-30
feat: add end event for piping

reconbot:
This will allow streams to end. We haven't done this because it allows a stream to be reopened and data to be continued to be read. This may not be a good idea but I'm opening this issue for discussion.

Maybe we should have this as an option?


HipsterBrown:
I agree that if this functionality is introduced, it should be off by default and enabled by an option in the constructor params. Otherwise I would count it in a major version bump.

HipsterBrown:
Looks good to me. 👍 :clap:

PR #2046, opened 2020-03-05
fix: reject on non-zero exit codes

Swaagie:
`udevadm` does not expose errors when exporting the database ([yet](https://github.com/systemd/systemd/pull/14960)). But it does exit with `code > 0`, if no errors are exposed the exit code can still be used to provide feedback even if generic rather than just returning an empty array.

Happy to change the thrown error message if it does not match requirements.

Background: https://github.com/systemd/systemd/issues/14959 wrong permissions on any device that `udevadm` attempts to list will throw `EACCES`. These errors are not shown but `strace` will show an exit code `1`

### DIFF
--- a/packages/bindings/lib/linux-list.js
+++ b/packages/bindings/lib/linux-list.js
@@ -42,7 +42,7 @@ function listLinux() {
     const ports = []
     const ude = childProcess.spawn('udevadm', ['info', '-e'])
     const lines = ude.stdout.pipe(new Readline())
-    ude.on('close', code => code && reject(new Error(`List exited with code: ${code}`)))
+    ude.on('close', code => code && reject(new Error(`Error listing ports udevadm exited with error code: ${code}`)))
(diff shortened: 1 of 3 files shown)


packages/bindings/lib/linux-list.js
```diff
@@ -42,6 +42,7 @@ function listLinux() {
     const ports = []
     const ude = childProcess.spawn('udevadm', ['info', '-e'])
     const lines = ude.stdout.pipe(new Readline())
+    ude.on('close', code => code && reject(new Error(`List exited with code: ${code}`)))
```
reconbot: I would say something like `Error listing ports udevadm exited with error code: ${code}`

Swaagie:
Updated, wasn't obvious at all that `udevadm` failed. I think on most systems it just always works. I just happened to have a pi3 that was being an outlier (read: I failed to set it up correctly 😄)

PR #2094, opened 2020-06-08
Add preferUnplugged to keep the main package plugged in Yarn 2

alecmev:
The `postinstall` causes Yarn to assume that the package might have a build step and Yarn ends up unplugging it, [which is undesirable](https://yarnpkg.com/advanced/lifecycle-scripts#a-note-about-postinstall). `preferUnplugged` is documented [here](https://yarnpkg.com/configuration/manifest#preferUnplugged).

### DIFF
--- a/packages/serialport/package.json
+++ b/packages/serialport/package.json
@@ -63,5 +63,6 @@
   "scripts": {
     "repl": "node bin/repl.js",
     "postinstall": "node thank-you.js"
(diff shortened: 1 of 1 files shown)


alecmev:
`@serialport/bindings`, indeed, needs to remain unplugged, but that doesn't concern the packages that depend on it.

PR #2102, opened 2020-06-18
Added low_latency mode for linux

jeremy-mf:
Summary:
On Linux systems, the USB latency timer can be reduced from ~16ms to ~1ms when the `ASYNC_LOW_LATENCY` flag is set on the `serial_struct`. This commit enables enabling and disabling this mode.
This feature mimics using `setserial /dev/tty[...] low_latency` from the command line, and the commit to PySerial here: https://github.com/pyserial/pyserial/pull/290

Tests:
- Updated the automated tests to account for the new flag
- Compiled on both Linux and Mac OS platforms
- Verified automated tests still pass and tested against an Arduino Uno
- Tested manually enabling and disabling this flag

### DIFF
--- a/packages/bindings/src/serialport_unix.cpp
+++ b/packages/bindings/src/serialport_unix.cpp
@@ -332,6 +332,17 @@ void EIO_Set(uv_work_t* req) {
     bits |= TIOCM_DSR;
   }
 
+  #if defined(__linux__)
+  int err = linuxSetLowLatencyMode(data->fd, data->low_latency);
+  if (err == -1) {
+    snprintf(data->errorString, sizeof(data->errorString), "Error: %s, cannot get", strerror(errno));
+    return;
+  } else if(err == -2) {
+    snprintf(data->errorString, sizeof(data->errorString), "Error: %s, cannot set", strerror(errno));
+    return;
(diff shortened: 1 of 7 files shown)


jeremy-mf:
For those it affects I think it will be a handy feature. Would love to get a review on the changes

ristomatti:
It doesn't seem sensible to automatically mark pull requests as stale "issues"...

mf-pherlihy:
Any update on this?

Shanmax1996:
How it enable😍

mf-pherlihy:
Can we get this merged?

packages/bindings/src/serialport_unix.cpp
```diff
@@ -359,6 +370,16 @@ void EIO_Get(uv_work_t* req) {
     return;
   }
 
+  data->lowLatency = false;
+  #if defined(__linux__) && defined(ASYNC_LOW_LATENCY)
```
reconbot, line 374: Why does this need the def check on `ASYNC_LOW_LATENCY`? And if it does need it do we need it elsewhere?

reconbot (approved):

Issue #2159, opened 2020-10-24
Serialport SHOULD support the latest version of Electron.


timothyjrbeck:
### What feature you'd like to see
(A clear and concise description of what the proposal is.)

Support for the latest version of Electron.

## Motivation
(Please outline the motivation for the proposal. It's interesting knowing what people are working on and also could help community members make suggestions for work-arounds until the feature is built)

As per the error message below, It seems like it's impossible to use the latest version of Electron (10.x) and serialport at the same time.

> Error: The module '\node_modules\@serialport\bindings\build\Release\bindings.node'
> was compiled against a different Node.js version using
> NODE_MODULE_VERSION 64. This version of Node.js requires
> NODE_MODULE_VERSION 84. Please try re-compiling or re-installing

Not using the latest version might cause some security issues, hence I think this problem should be addressed as soon as possible.

reconbot:
Patches welcome, it usually involves updating prebuild.

Lovinity:
It is possible to use node-serialport and the latest Electron if you build from source. Unfortunately, node-serialport does not support source building on MacOS (at least, that's the error I get). I would love to at the very least see support for source building on MacOS.

It is very critical to support later versions of Electron soon. Given the latest binary-available version is electron v9, once electron v12 is out, electron v9 will no longer be supported by the Electron team.

reconbot:
It absolutely does support building locally.

Lovinity:
That's odd. When I tried to compile on OSX 10.2 it said building wasn't supported on that architecture.

reconbot:
[Github uses](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11.0-Readme.md)
> System Version: macOS 11.0.1 (20B29)

And I use 
> 10.15.7

but nothing I know about has changed substantially since v8 iirc

Lovinity:
I'll see if I can bump the build server up to macOS 11 and see if that resolves the problem

Lovinity:
Okay, I got it backwards. MacOS was the only architecture that supported compiling from source. It failed both on linux and on win32.

The most likely reason for this is the build environment itself runs in MacOS... it has to because you cannot compile anything MacOS on another architecture because of their proprietary XCode lockdown. I am going to try and split the build environment into multiple architectures and see if that fixes the issue.

reconbot:
[...]

Issue #2156, opened 2020-10-20
Node ABI issue with 9.0.2 bindings


choppsta:
### Summary of Problem

- What are you trying to do?

Install serialport with prebuilt bindings

- What happens?

They cannot be found

- What should have happened?

They should be matched against the node ABI number and downloaded

### Versions, Operating System and Hardware

- SerialPort@9.0.2
- Node.js 12+
- Windows, Linux, Mac

It looks like the process that builds the prebuilt bindings is affected by this issue:
https://github.com/lgeiger/node-abi/issues/90

The filenames of the bindings don't seem to have the right ABI numbers (e.g. node 12 should be 72, but I believe it's incorrectly being labelled as 68).

It would also be really great to get prebuilt bindings for electron 10 (ABI 82)!

reconbot:
Until that issue closes there's not much we can do.

Lovinity:
It looks like node-abi released an update a few days ago.

jeffrson:
Also there are'nt binary files for node 14 (LTS!) ABI 83.

reconbot:
Try `serialport@9.0.3`

reconbot:
And as always I welcome help maintaining the project, I'm not able to attend to these changes promptly.

choppsta:
Thanks reconbot, unfortunately I think there is still an issue with the bindings build process as the ABI numbers for the node bindings don't seem to match (https://github.com/lgeiger/node-abi/blob/master/abi_registry.json).  The electron ones are ok, although they only go up to 80 (Electron 9) where as the current version of Electron is two major versions ahead now.  A lot of the recent issues on this repo seem to be related to this.

I really appreciate all the work you put into this project and would love to help out if I can.  Could you point me in the right direction for how the building of these bindings works?

choppsta:
Hi reconbot, I've dug into this and found the source of the issue:

The version of the `node-abi` in the `package-lock.json` is out of date.  It's version `2.19.1` when the current version is `2.19.3`.  Because it's a dependency of `prebuild` it won't get updated unless that project specifies a higher version.

Quick solution is `npm update --depth 2 node-abi`, although it might be better to add `node-abi` as a dependency to this project meaning that it could be kept up to date more easily with `npm update`.

Hope this helps.  Let me know I can do anything to help further.

jeffrson:
Good find @choppsta!

[...]

Issue #2140, opened 2020-09-16
Support Electron 10


panuhorsmalahti:
### What feature you'd like to see

Support for Electron 10.x.

## Motivation

node-serialport should support latest version of Electron.

This can likely be accomplished simply by rebuilding and releasing this package.

JuliadsonS:
Alguma novidade? funciona perfeitamente com a versão 8.2.0 do eléctron!

ukasiu:
this will never become stale :)

mkurczewski:
still nothing?

RHeynsZa:
Any workaround? Other than downgrading electron

orange-beans:
Electron 10 and Electron 11 not working.

reconbot:
I released a serialport with the latest abi support a few days ago.

orange-beans:
> I released a serialport with the latest abi support a few days ago.

Thank you for the reply. The latest binding I can find for v9.0.3 is v80 (bindings-v9.0.3-electron-v80-win32-x64.tar.gz);
However, the error message I encountered said it requires NODE_VERSION_V85.

My configuration is as following:
Electron: v11.0.3
Node:v12.16.1
serialport: v9.0.3

Anything I'm missing?

Thanks with regards.

reconbot:
Lets try it again with 9.0.4

reconbot:
Looks like we're having build issues on linux and windows. I'm digging into it.

reconbot:
I've ported the builds to github actions, seems to be in good shape.

orange-beans:
> I've ported the builds to github actions, seems to be in good shape.

Thanks for the speedy update! Now the Electron App can successfully be loaded with _serialport_ binding.

However, a new problem has been encountered as shown below. (which has never been seen in version 8.x.x)

The configuration:
Electron: v11.0.3
Node:v12.16.1
serialport: v9.0.3 (Binding version V85)

I tried to create a new port with following code:
```Javascript
const testPort = new SerialPort('COM1', defaultPortConfig, function(err) {
  if (err) console.log(err);
  else {
    console.log('Sucess')
  }
});
```

And the following error pops out:
```Javascript
TypeError: Third argument must be a function
    at internal/util.js:297:30
    at new Promise (<anonymous>)
    at open (internal/util.js:296:12)
[...]

PR #2190, opened 2020-12-16
Explicitly specify node-abi version in top-level package.json

NoahAndrews:
[As was discovered](https://github.com/serialport/node-serialport/issues/2156#issuecomment-738986150) by @choppsta, the node-abi also needs to be specified in the top-level package.json. This PR will hopefully accomplish what #2183 was meant to accomplish.

As a side note, I believe this wouldn't have been an issue if this library was using lerna's hoisting option. Enabling that is something that should probably be considered.

If this works, #2140, #2156, and #2159 should be closed.

### DIFF
--- a/packages/bindings/package-lock.json
+++ b/packages/bindings/package-lock.json
@@ -4,37 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@serialport/binding-abstract": {
-      "version": "9.0.2",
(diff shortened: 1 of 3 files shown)


packages/bindings/package-lock.json
```diff
@@ -4,37 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@serialport/binding-abstract": {
```
reconbot, line 7: unrelated change?

NoahAndrews: I don't know. I ran `npm install`, and this is one of the things it did.

PR #2197, opened 2021-01-04
Add packet length parser

### LINKED ISSUE #1609: Make a parser to pull data out from the middle of a known binary packet
We don't have a parser that makes this [very common problem](https://github.com/node-serialport/node-serialport/issues/1506#issuecomment-371834751) easy.

doggkruse:
Add a parser for data that has a delimiter and a length in the byte stream.  This should resolve #1609

### DIFF
--- a/packages/parser-packet-length/README.md
+++ b/packages/parser-packet-length/README.md
@@ -1,0 +1,1 @@
+See our api docs https://serialport.io/docs/api-parser-packet-length
--- a/packages/parser-packet-length/lib/index.js
+++ b/packages/parser-packet-length/lib/index.js
@@ -1,0 +1,79 @@
+const { Transform } = require('stream')
+
+/**
+* A transform stream that decodes packets with a delimiter and length of payload
+* specified within the data stream.
+* @extends Transform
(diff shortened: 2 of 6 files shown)


packages/parser-packet-length/README.md
```diff
@@ -0,0 +1 @@
+See our api docs https://serialport.io/docs/api-parser-packet-length
```
reconbot, line 1: We don't have docs here

packages/parser-packet-length/lib/index.js
```diff
@@ -0,0 +1,79 @@
...
+  length_offset: 2,
+}))
+*/
+class PacketLengthParser extends Transform {
+  constructor(options = {}) {
+    super(options)
+
+    const opts = {
+      delimiter: 0xaa,
+      packet_overhead: 2,
```
reconbot, line 33: Please use camel case keys eg `packetOverhead`

reconbot (commented):
Just a few nits, looks pretty good!

doggkruse:
Commets should be addressed in a28dc87

reconbot (approved):

Issue #2240, opened 2021-04-29
serialport.set fails on Linux due to Low Latency Mode (@serialport/bindings v9.0.7)


GazHank:
### Summary of Problem

- What are you trying to do? set({dtr: true}) or set({dtr: false}) 
- What happens? Error: Operation not supported, cannot set (thrown by serialport_unix.cpp line 342)
- What should have happened? operation should succeed without error

### Code to Reproduce the Issue

```js
const serialport = new SerialPort(path)
const delay = time => new Promise(res=>setTimeout(res,time))

serialport.on('open', function() {
    serialport.set({dtr: true});
    await delay(1000)
    serialport.set({dtr: false});
}

```

### Versions, Operating System and Hardware

- SerialPort v9.0.7 (@serialport/bindings v9.0.7)
- Node.js v14.16.0
- Linux 20.04.2 LTS 64bit 
- Hardware and chipset? (attempting to send dtr to arduino:samd)

### Additional info
This issue replaces issue https://github.com/serialport/node-serialport/issues/2239 to improve clarity.

It appears that this issue relates to the implementation of https://github.com/serialport/node-serialport/pull/2102. Specifically within the set method at [serialport_unix.cpp (line 335)](https://github.com/jeremy-mf/node-serialport/blob/56caf5ea1299cd62483105fabcc4fe25a7c4daee/packages/bindings/src/serialport_unix.cpp#L335) there is an attempt made to set low latency mode irrespective of if lowLatency was specified in the method call.  Adding a check for ```  if (data->lowLatency)```  per below seems to resolve this, but there may be another more preferable solution.

```  
#if defined(__linux__)
  if (data->lowLatency) {
    int err = linuxSetLowLatencyMode(data->fd, data->lowLatency);
    if (err == -1) {
      snprintf(data->errorString, sizeof(data->errorString), "Error: %s, cannot get", strerror(errno));
      return;
    } else if(err == -2) {
      snprintf(data->errorString, sizeof(data->errorString), "Error: %s, cannot set", strerror(errno));
      return;
    }
  }
  #endif
```

I have tried adding a check for `defined(ASYNC_LOW_LATENCY)` but this doesnt seem to help...

This issue is not present in @serialport/bindings v9.0.4, however if serialport is defined in the dependencies, then when the packages are installed the underlying packages (e.g. bindings) will be upgraded to v9.0.7 resulting in this issue appearing for anyone using serialport >=9.0.0

[...]

PR #2241, opened 2021-04-29
Linux low latency fix (allow set without changing low latency mode)

GazHank:
Only set low Latency mode if lowLatency is defined

Proposed fix to address https://github.com/serialport/node-serialport/issues/2240

### DIFF
--- a/packages/bindings/src/serialport_linux.cpp
+++ b/packages/bindings/src/serialport_linux.cpp
@@ -43,6 +43,8 @@ int linuxSetLowLatencyMode(const int fd, const bool enable) {
   if (ioctl(fd, TIOCGSERIAL, &ss)) {
     return -1;
   }
+  
+  int oldFlags = ss.flags;
 
   if (enable) {
     ss.flags |= ASYNC_LOW_LATENCY;
@@ -50,9 +52,24 @@ int linuxSetLowLatencyMode(const int fd, const bool enable) {
     ss.flags &= ~ASYNC_LOW_LATENCY;
   }
 
-  if (ioctl(fd, TIOCSSERIAL, &ss)) {
-    return -2;
+  if (oldFlags != ss.flags)
+  {
+    if (ioctl(fd, TIOCSSERIAL, &ss)) {
+      return -2;
+    }
   }
+
+  return 0;
+}
+
+int linuxGetLowLatencyMode(const int fd, bool* const enabled) {
+  struct serial_struct ss;
+
+  if (ioctl(fd, TIOCGSERIAL, &ss)) {
+    return -1;
+  }
+
+  *enabled = ss.flags & ASYNC_LOW_LATENCY;
 
   return 0;
 }
--- a/packages/bindings/src/serialport_linux.cpp
+++ b/packages/bindings/src/serialport_linux.cpp
@@ -17,7 +17,7 @@ int linuxSetCustomBaudRate(const int fd, const unsigned int baudrate) {
     t.c_cflag |= BOTHER;
     t.c_ospeed = t.c_ispeed = baudrate;
 
-    if (ioctl(fd, TCSETS2, &t)) {
+    if (ioctl(fd, TCSETS2, &t) < 0) {
       return -2;
     }
 
@@ -43,22 +43,19 @@ int linuxSetLowLatencyMode(const int fd, const bool enable) {
   if (ioctl(fd, TIOCGSERIAL, &ss)) {
     return -1;
   }
-  
-  int oldFlags = ss.flags;
 
-  if (enable) {
-    ss.flags |= ASYNC_LOW_LATENCY;
-  } else {
-    ss.flags &= ~ASYNC_LOW_LATENCY;
-  }
+  if (ss.flags & ASYNC_LOW_LATENCY != enable) {
+    
+    if (enable) {
+      ss.flags |= ASYNC_LOW_LATENCY;
+    } else {
+      ss.flags &= ~ASYNC_LOW_LATENCY;
+    }
 
-  if (oldFlags != ss.flags)
-  {
-    if (ioctl(fd, TIOCSSERIAL, &ss)) {
+    if (ioctl(fd, TIOCSSERIAL, &ss) < 0) {
       return -2;
     }
   }
-
   return 0;
 }
 
--- a/packages/bindings/src/serialport_unix.cpp
+++ b/packages/bindings/src/serialport_unix.cpp
@@ -333,13 +333,15 @@ void EIO_Set(uv_work_t* req) {
   }
 
   #if defined(__linux__)
-  int err = linuxSetLowLatencyMode(data->fd, data->lowLatency);
-  if (err == -1) {
-    snprintf(data->errorString, sizeof(data->errorString), "Error: %s, cannot get", strerror(errno));
-    return;
-  } else if(err == -2) {
-    snprintf(data->errorString, sizeof(data->errorString), "Error: %s, cannot set", strerror(errno));
-    return;
+  if (data->lowLatency) {
+    int err = linuxSetLowLatencyMode(data->fd, data->lowLatency);
+    if (err == -1) {
+      snprintf(data->errorString, sizeof(data->errorString), "Error: %s, cannot get", strerror(errno));
+      return;
+    } else if(err == -2) {
+      snprintf(data->errorString, sizeof(data->errorString), "Error: %s, cannot set", strerror(errno));
+      return;
+    }
   }
   #endif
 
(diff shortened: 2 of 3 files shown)


packages/bindings/src/serialport_unix.cpp
```diff
@@ -333,13 +333,15 @@ void EIO_Set(uv_work_t* req) {
...
 
   #if defined(__linux__)
-  int err = linuxSetLowLatencyMode(data->fd, data->lowLatency);
-  if (err == -1) {
-    snprintf(data->errorString, sizeof(data->errorString), "Error: %s, cannot get", strerror(errno));
-    return;
-  } else if(err == -2) {
-    snprintf(data->errorString, sizeof(data->errorString), "Error: %s, cannot set", strerror(errno));
-    return;
+  if (data->lowLatency) {
```
reconbot, line 336: so if it's undefined or false its not set, and if it's true it is set? does this mean you can't turn it off? Would you ever want to turn it off?

GazHank: Hi @reconbot, thanks for the quick response :-)

I have to admit, toggling low latency on and off isn't really something I considered.  Given how the user sets low latency within linux, and the use cases I've seen I don't know if toggling is common.

I'm not certain this is the best fix, and bigger change may be needed to fully address this feature, but my priorities when proposing this were:
1. to get the set method to work again for people who don't need to invoke low latency mode
2. continue to allow anyone who does want to invoke the low latency mode to enable it
3. keep the change as small and low impact as possible (so we can fix point 1 asap)

As I understand it, the current code and the default behaviour of Linux are not aligned.  In particular, I believe that in order to be able to invoke lowLatency mode you need to first change the default behaviour of the port, using something like: `sudo setserial /dev/ttyUSB0 low_latency`.  As such, by default Linux will not allow you to invoke low latency mode in Serialport, however the code within the set method tries to configure configure it irrespective of the options the user supplies to the set method.  Even specifying `port.set({dtr: true, LowLatency: false })` or `port.set({dtr: true})` will cause an attempt to set the low latency bit to zero and error out.

Having thought about this a little more, I wonder if the set method is really the right place to be activating low latency mode, since all of the other options within the set method are really control signals.  I can't help thinking about low latency mode being more like setting the baud rate etc.  As such it would seem to me to be more appropriate to allow it to be defined within the port open method, and within the update method.  Of course we probably need to find a way to check if we have permission to enable low latency mode first otherwise we will continue to throw errors.  The potential downside of treating low latency mode like baud rate is that baud rate has it's own get method, it would be a shame to have to add a get method just for this since it is platform specific.

Let me know if you would like me to try a different way to fix this

Cheers

Gareth

GazHank: Upon further investigation I think we need a better fix for this.  I'll add more info to the issue based on my improved understanding of what's going on...

I will try to make a new PR to address this more completely ASAP, but until a better fix is available I still think this is an acceptable workaround to get the v9 code working on Linux for all users.

Per my comment in the issue, anyone placing a dependency on "serialport": "9.0.0" will get the broken version of bindings since the interdependence is specified as "^9.0.0".

GazHank:
Improved fix for low latency mode:

### Changes
serialport_linux.cpp:
- Revise the linux set method to only attempt an update if required (if the bit doesn't need to be flipped then don't attempt an unnecessary call for which we might not have permission)
- Addition of a linux specific linuxGetLowLatencyMode method - for consistency with linuxGetSystemBaudRate, allowing removal of the logic from the more general unix code

serialport_linux.h:
- Addition of a linux specific linuxGetLowLatencyMode method - for consistency with linuxGetSystemBaudRate

serialport_unix.cpp:

- Revise sequence of set method - to allow control bits (e.g. dtr) to be set even if low latency fails
- Revise sequence of get method - to allow control bits (e.g. dtr) to be returned even if get of low latency info fails
- Revise get method to invoke linux specific method (point 2 above, and removing broken logic in the process)
- Revise the error messages for the set and get messages for the low latency calls to make it clearer when these fail - since they are likely to fail when we don't have suitable access rights

### Testing
Tested on Linux Ubuntu LTS and Win10:

**Linux without admin rights**
- [x] Set method call {dtr: true} - succeeds without error
- [x] Set method call {dtr: true, lowLatency: true} - dtr correctly set, error reported for lowLatency, but handled without issue
- [x] Get method call - succeeds without error

**Linux with admin rights**
- [x] Set method call {dtr: true} - succeeds without error
- [x] Set method call {dtr: true, lowLatency: true} - succeeds without error
- [x] Get method call - succeeds without error

**Windows 10**
- [x] Set method call {dtr: true} - succeeds without error
- [x] Set method call {dtr: true, lowLatency: true} - succeeds without error (lowLatency ignored)
- [x] Get method call - succeeds without error

### Additional considerations
The the focus of this change was entirely on addressing errors per serialport#2240 as such I have not performed any actual benchmarking of the lowLatency mode to check for changes to latency (or increases to CPU load - [setserial](https://linux.die.net/man/8/setserial) has info on expected benefits/impacts).

GazHank:
@reconbot do let me know if you would like to discuss or revise any of these changes, or if you would prefer if I setup a new PR for this

packages/bindings/src/serialport_linux.cpp
```diff
@@ -43,17 +43,34 @@ int linuxSetLowLatencyMode(const int fd, const bool enable) {
...
     ss.flags |= ASYNC_LOW_LATENCY;
   } else {
     ss.flags &= ~ASYNC_LOW_LATENCY;
   }
 
-  if (ioctl(fd, TIOCSSERIAL, &ss)) {
-    return -2;
+  if (oldFlags != ss.flags)
+  {
+    if (ioctl(fd, TIOCSSERIAL, &ss)) {
```
reconbot, line 57: I think this is wrong - https://man7.org/linux/man-pages/man2/ioctl.2.html#RETURN_VALUE

It should be a positive number on success not failure.

packages/bindings/src/serialport_linux.cpp
```diff
@@ -43,17 +43,34 @@ int linuxSetLowLatencyMode(const int fd, const bool enable) {
   if (ioctl(fd, TIOCGSERIAL, &ss)) {
     return -1;
   }
+  
+  int oldFlags = ss.flags;
```
reconbot, line 47: lets see if `oldFlags & ASYNC_LOW_LATENCY == enable` and just return early if we don't have to do anything?

packages/bindings/src/serialport_unix.cpp
```diff
@@ -332,17 +332,6 @@ void EIO_Set(uv_work_t* req) {
     bits |= TIOCM_DSR;
   }
 
-  #if defined(__linux__)
```
reconbot, line 335: good move

reconbot (commented):
Sorry for the delay in review. I think you uncovered a bug, and also some feedback on simplifying the call to set.

packages/bindings/src/serialport_linux.cpp
GazHank: good shout that is lots nicer

packages/bindings/src/serialport_linux.cpp
GazHank: agreed, looks like we should check for negative values to indicate errors instead:
`if (ioctl(fd, TIOCSSERIAL, &ss) < 0) {`

although the same also seems to be true for the setting the baud rate:
`    if (ioctl(fd, TCSETS2, &t)) {
      return -2;
    }`
does that need a fix too?

packages/bindings/src/serialport_linux.cpp
```diff
@@ -43,22 +43,19 @@ int linuxSetLowLatencyMode(const int fd, const bool enable) {
...
-  
-  int oldFlags = ss.flags;
 
-  if (enable) {
-    ss.flags |= ASYNC_LOW_LATENCY;
-  } else {
-    ss.flags &= ~ASYNC_LOW_LATENCY;
-  }
+  if (ss.flags & ASYNC_LOW_LATENCY != enable) {
+
```
reconbot, line 48: just a nitpick it's better to return early than it is to nest

packages/bindings/src/serialport_linux.cpp
```diff
@@ -43,22 +43,19 @@ int linuxSetLowLatencyMode(const int fd, const bool enable) {
...
+    if (enable) {
+      ss.flags |= ASYNC_LOW_LATENCY;
+    } else {
+      ss.flags &= ~ASYNC_LOW_LATENCY;
+    }
 
-  if (oldFlags != ss.flags)
-  {
-    if (ioctl(fd, TIOCSSERIAL, &ss)) {
+    if (ioctl(fd, TIOCSSERIAL, &ss) < 0) {
```
reconbot, line 55: another nitpick ioctl will always return -1 on error never anything else

reconbot (approved):

PR #2253, opened 2021-05-26
fix: linux baudRate and latency errors

reconbot:
- linuxGetLowLatencyMode would incorrectly get an error
- linuxSetLowLatencyMode would incorrectly get an error
- linuxGetSystemBaudRate would incorrectly get an error
- linuxSetCustomBaudRate would incorrectly get an error

Thanks to @GazHank for highlighting work was needed in this area

### DIFF
--- a/packages/bindings/src/serialport_linux.cpp
+++ b/packages/bindings/src/serialport_linux.cpp
@@ -9,7 +9,7 @@
 int linuxSetCustomBaudRate(const int fd, const unsigned int baudrate) {
     struct termios2 t;
 
-    if (ioctl(fd, TCGETS2, &t)) {
+    if (ioctl(fd, TCGETS2, &t) == -1) {
       return -1;
     }
 
@@ -17,7 +17,7 @@ int linuxSetCustomBaudRate(const int fd, const unsigned int baudrate) {
     t.c_cflag |= BOTHER;
     t.c_ospeed = t.c_ispeed = baudrate;
 
-    if (ioctl(fd, TCSETS2, &t) < 0) {
+    if (ioctl(fd, TCSETS2, &t) == -1) {
       return -2;
     }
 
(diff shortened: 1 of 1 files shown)


packages/bindings/src/serialport_linux.cpp
```diff
@@ -40,29 +40,31 @@ int linuxGetSystemBaudRate(const int fd, int* const outbaud) {
...
   }
 
-  if (ss.flags & ASYNC_LOW_LATENCY != enable) {
-    
-    if (enable) {
-      ss.flags |= ASYNC_LOW_LATENCY;
-    } else {
-      ss.flags &= ~ASYNC_LOW_LATENCY;
-    }
+  if (ss.flags & ASYNC_LOW_LATENCY == enable) {
```
GazHank, line 47: Not entirely certain why... but the compiler doesn't seem to like this recent change.  At first I thought this was a rogue warning, but it is actually misreading the code if we dont add parentheses around the items per below:
`if ((ss.flags & ASYNC_LOW_LATENCY) == enable) {`

the warnings are:
```
./src/serialport_linux.cpp: In function ‘int linuxSetLowLatencyMode(int, bool)’:
../src/serialport_linux.cpp:47:36: warning: comparison of constant ‘8192’ with boolean expression is always false [-Wbool-compare]
   47 |   if (ss.flags & ASYNC_LOW_LATENCY == enable) {
      |                                    ^
../src/serialport_linux.cpp:47:36: warning: suggest parentheses around comparison in operand of ‘&’ [-Wparentheses]

```
Without adding the parentheses the compiled code does try to set the low latency flag every time, irrespective of the bool enable value that we pass

I guess my c is a little rustier that I thought

GazHank:
With the parentheses change per above, I'm getting successful results:
- setting DTR true etc seems to work
- no errors when setting low latency with admin rights
- suitable error messages when access rights don't permit setting low latency

I've tested on:
- Ubuntu 20.04.2 LTS 
- Node.js 14.16.0, 
- Chromium 89.0.4389.128, 
- Electron 12.0.9

I should call out that I'm not really able to test that low latency mode works, only that it doesn't seem to brake stuff anymore :-)

packages/bindings/src/serialport_linux.cpp
reconbot: I love how it does compile anyway

GazHank:
@reconbot let me know if there are any extra tests you'd like me to run.  I'm normally on Win10, but can fire up Linux for some testing if I can help

PR #2267, opened 2021-06-14
Issue template: Request extra info about Electron, Prebuilds and System Architecture

GazHank:
Just a small change to ask for info about the environment to aid trouble shooting.

Not sure if we need to give more info or links so that people know how to answer the extra questions, but I hope this extra info it could help cut down some of the troubleshooting steps


reconbot (approved):
I wish people filled it out more often but this doesn't hurt one bit.

PR #2278, opened 2021-06-20
Ignore end of line prettier errors

GazHank:
Override the default lint rule for End of Line introduced in Prettier v2.0.0 (i.e. revert to "auto" rather than forcing "LF")

While there are other options to address this, this change simply suppresses the "error" rather than touching every single line within 70+ files

See [prettier options](https://prettier.io/docs/en/options.html#end-of-line) for more info

Perhaps we could find a good suitable time to switch most/all of the linting rules over to the default, but I fear it will make it quite messy to trace the history of commits if a lint change touches every line in every file...

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -47,7 +47,8 @@
         "trailingComma": "es5",
         "semi": false,
         "printWidth": 150,
-        "arrowParens": "avoid"
+        "arrowParens": "avoid",
+        "endOfLine": "auto"
       }
     ],
     "mocha/no-exclusive-tests": "error",


GazHank:
`npm run lint` generates a lot of noise.  For example:

```
C:\Git\node-serialport\test\initializers\test-config.js
   1:53  error  Delete `␍`  prettier/prettier
   2:1   error  Delete `␍`  prettier/prettier
   3:61  error  Delete `␍`  prettier/prettier
   4:63  error  Delete `␍`  prettier/prettier
   5:64  error  Delete `␍`  prettier/prettier
   6:37  error  Delete `␍`  prettier/prettier
   7:85  error  Delete `␍`  prettier/prettier
   8:6   error  Delete `␍`  prettier/prettier
   9:30  error  Delete `␍`  prettier/prettier
  10:4   error  Delete `␍`  prettier/prettier
  11:2   error  Delete `␍`  prettier/prettier

✖ 6990 problems (6990 errors, 0 warnings)
  6990 errors and 0 warnings potentially fixable with the `--fix` option.
```

I'm pretty sure this isnt down to the way I've cloned the repository on Windows as I don't get these errors for other reps I use.  But if this i just a problem for me then I'll see if I can change my local config

GazHank:
I wonder if my windows machine defaulted to some old school setting when I did a complete rebuild a few weeks ago :-(

I hadn't checked the linting on this repo before so maybe hadn't noticed.  Hopefully this is just a local config problem :-)

GazHank:
Sorry it looks like a local config issue from the rebuild

fixed with:
```
git config --global core.autocrlf false
git add --renormalize .
```

I'll close this, and see if the Lerna update needs to be reworked

PR #2294, opened 2021-07-06
Use Dependabot to keep node-abi up to date

GazHank:
This is an attempt to stay on top of updates to node and electron versions as they are released.  We could use Dependabot for more updates, but I thought it safest to try this out with a narrow scope for now

Proposed per discussion https://github.com/serialport/node-serialport/discussions/2289

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,32 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+    allow:
+      # Allow updates for node-abi
+      - dependency-name: "node-abi"
+    # Prefix all commit messages with "Composer"
+    # include a list of updated dependencies
+    commit-message:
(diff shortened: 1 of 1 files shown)


GazHank:
I believe the reduction in codecov is simply due to the introduction of a config file rather than actual code, so nothing to worry about.

GazHank:
I've added this to my fork of the repo (which had an old version of node-abi) and it seems to work as expected:

https://github.com/GazHank/node-serialport/pulls

GazHank:
Hmm, I'm going to double check if we can just run it on the root, or if we need to explicitly run it per package.json

GazHank:
@reconbot keen to make sure you are happy before this is included.  I don't want to generate unnecessary noise so this should only propose updates to the single package for now.  If we want to expand this in future we could include all dependencies, just direct ones or just those direct dependencies that are used for production.

Having activated the bot for all modules in my local fork I think we might want to check this is providing value before we allow this to do too much for us

GazHank:
Ok, so let's try this for now, and if the bot behaves we can see about expanding its remit

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+    allow:
+      # Allow updates for node-abi
+      - dependency-name: "node-abi"
+    # Prefix all commit messages with "Composer"
+    # include a list of updated dependencies
+    commit-message:
+      prefix: "Chore: "
```
reconbot, line 18: Lowercase chore, this might work as is

PR #2298, opened 2021-07-19
Automate changelog

GazHank:
Initial implementation of lerna changelog

Changelog labels configured to match previous changelog entries (with a little extra :boom:)

_Based on a quick check, as long as PRs are (re)label before the publication, the items apear in the right group...I haven't checked what happens if items have multiple labels (yet)_

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "eslint-plugin-prettier": "^3.3.1",
     "eslint-plugin-promise": "^4.3.1",
     "lerna": "^3.22.1",
+    "lerna-changelog": "^1.0.1",
     "mocha": "^8.3.0",
     "node-abi": "^3.0.0",
     "nyc": "^15.1.0",
@@ -47,5 +48,14 @@
     "proxyquire": "^2.1.3",
     "sinon": "^9.2.4"
   },
-  "funding": "https://opencollective.com/serialport/donate"
+  "funding": "https://opencollective.com/serialport/donate",
+  "changelog": {
+    "labels": {
+      "breaking": ":boom: BREAKING CHANGES :boom:",
+      "feature-request": "Features",
+      "bug": "Bug Fixes",
+      "docs": "Documentation",
+      "internal": "Chores"
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -47,5 +48,14 @@
...
   },
-  "funding": "https://opencollective.com/serialport/donate"
+  "funding": "https://opencollective.com/serialport/donate",
+  "changelog": {
+    "labels": {
+      "breaking": ":boom: BREAKING CHANGES :boom:",
+      "feature-request": "Features",
+      "bug": "Bug Fixes",
+      "docs": "Documentation",
+      "internal": "Chores"
```
reconbot, line 58: I use `chore:` a lot, that still work? These are commit prefix's right?

reconbot (approved):
looks good for me - should we delete or merge all the old files?

package.json
GazHank: They are GitHub labels... the Lerna changelog seems to be based on PR labels in GitHub, rather than commit prefixes

So if we set the label as "internal" then they are grouped under "chore" in the changelog.

Since I can't seem to get it to work with historic records there is only one PR to play with for now... (you can change the label etc for that PR and see how it shows in the log).

Ps - If you think it is clearer then we could rename the "internal" label as "chore" for consistency

GazHank:
I think we are safe to add this, but I would like to check it works for the next release or two before we delete anything

GazHank:
Does it list the changelog for you?  I wonder if we need to specifically define the repo and cacheDir (rather than let it be inferred per https://github.com/lerna/lerna-changelog) Gareth Hancock+44 7789 323231 From: Francis GulottaSent: 28 September 2021 19:33To: serialport/node-serialportCc: Gareth Hancock; AuthorSubject: Re: [serialport/node-serialport] Automate changelog (#2298) got this on publish? Are you sure you want to publish these packages? Yeslerna info execute Skipping releaseslerna info git Pushing tags...lerna info publish Publishing packages to npm...lerna info Verifying npm credentials—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.

GazHank:
sorry I'll have to take another look at this, I haven't really touched it since getting stuck in the NAPI stuff

Issue #2191, opened 2020-12-19
TypeError: Third argument must be a function


Lovinity:
### Summary of Problem

(Please answer all 3)

- What are you trying to do?: Open a connection to a serial port device via the serialport constructor on Windows 10, electron 11.1, bindings 9.0.4.
- What happens? When attempting to establish a connection to a serialport, it throws an error saying third argument must be a function.
- ``` TypeError: Third argument must be a function
    at internal/util.js:297:30
    at new Promise (<anonymous>)
    at open (internal/util.js:296:12)
    at WindowsBinding.open (C:\Users\user\wwsu-dj-controls\node_modules\@serialport\bindings\lib\win32.js:56:22)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) ```
- What should have happened? The connection establishes successfully and we are able to begin writing (which was the case in electron 9 and bindings 9.0.2).

### Code to Reproduce the Issue

```js

// let device = device path retrieved from electron.settings, which I verified is operating correctly.

delaySerial = new serialport(device, {
	baudRate: 38400,
});

delaySerial.on("error", (err) => {
	serialError(err); // function sends the error both to the console and another logging utility. It also disconnects the serialport if not already disconnected and attempts to reconnect after a 15 second delay.
});
```

### Versions, Operating System and Hardware

- SerialPort@? 9.0.4
- Node.js v? Electron 11.1.0
- Windows? Linux? Mac? Windows 10 build 2004
- Hardware and chipset? (Prolific/FTDI/Other) Unknown/other (uses USB connection, but this error occurs no matter which COM port I try)

orange-beans:
@Lovinity Encountered same problem. This problem was not found in serialport@9.0.3 + electron@8 though.

pklaschka:
Just for reference (this isn't really a solution, of course, if anything, it's an ugly hack): After encountering the same error, I was able to get it to work with Electron 11.1.0 and serialport in version 8.0.8 (after running `electron-rebuild`, of course) on both Windows and Linux.

Seeing

> @Lovinity Encountered same problem. This problem was not found in serialport@9.0.3 + electron@8 though.

might, of course, mean, that a downgrade to an exact 9.0.3 might also have solved that problem, of course (I probably didn't properly specify the exact version, which has then lead me to believe 9.0.3 wasn't working either 😅 ).

Lovinity:
[...]

PR #2305, opened 2021-08-06
Node-API migration

GazHank:
DRAFT PR - FOR REVIEW ONLY

This is an initial draft of the migration to NAPI, inaddition to the automated migration tool quite a lot of the migration has to be hand crafted.  So far only the core / architecture agnostic functionality has been updated.

I have tried to keep all changes:
- as small as possible
- replacing logic like for like
- applying the same changes throughout for consistency

As such some enhancement, improvements or simplifications that we gain from switching to NAPI may not be present (e.g. revised error handling)

**Changes:**

- [x] Automated conversion to NAPI

SerialPort.cpp
- [x] Update type conversions not handled automatically (we also need to decide if we retain int32 or just switch to int64 now)
- [x] String conversions updated
- [x] Add missing return types
- [x] Switch uv_work_t to napi_async_work (uv_queue_work to napi_queue_async_work)
- [x] Switch return argv (array) to args (vector)
- [x] Revise delete to napi_delete_async_work and free
- [x] Avoid uncheck napi_* calls for return args (per feedback from @KevinEady)
- [x] Replace napi_create_async_work with Napi::AsyncWorker (per feedback from @KevinEady and @NickNaso)

Serialport.h
- [x] Change main methods to return Napi::Value and consume const Napi::CallbackInfo& info
- [x] Change EIO methods to consume napi_env env, void* req
- [x] Change EIO_After methods to consume napi_env env, napi_status status, void* req
- [x] Remove workaround for Electron 11, as this version is no longer supported ~~Replace API version check with ABI stable runtime check~~ 
  Change structs:
  - [x] Remove all references to AsyncResource (currently commented out)
  - [x] Switch callback to use Napi::FunctionReference
  - [x] Include napi_async_work work attribute
  - [x] Replace AsyncResource with Napi::AsyncWorker (per feedback from @KevinEady and @NickNaso)

Serialport_win
- [x] Same updates as serialport.cpp
- [x] Replace napi_create_async_work with Napi::AsyncWorker (common methods)
- [x] Replace napi_create_async_work with Napi::AsyncWorker (list)
- [x] Replace uv_async_t with Napi::AsyncWorker (read & write)
- [x] Fix event handling (close / disconnect event)

poller
- [x] Automated conversion
- [x] Manual conversion
- [ ] fix destroy ()

darwin_list
- [x] Same updates as serialport.cpp
- [x] Replace napi_create_async_work with Napi::AsyncWorker (list)

binding.gyp 
- [x] Use include_dir instead of include. (per feedback from @NickNaso)

**Please note: This should now be ready for testing.  Please let me know if you encounter any issues or unexpected behaviour**


cinderblock:
What is needed to get this over the edge? Or is this possibly complete and just needs reviews and approval? I'd love to have this working or test with it.

.generators/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "main": "lib/{{dashCase name}}.js",
   "dependencies": {
+    "node-addon-api": "4.0.0",
```
reconbot, line 6: probably not needed here

packages/binding-abstract/package.json
```diff
@@ -6,6 +6,7 @@
     "serialport-binding"
   ],
   "dependencies": {
+    "node-addon-api": "4.0.0",
```
reconbot, line 9: or here - as there are no binary anything in this package

reconbot (commented):
I am not qualified to review this for c++ correctness but the logic doesn't seem to have changed. This is amazing.

You'll only need the napi package in the bindings package, you can leave the others alone.

I think we'll have to take a look at what prebuild can do to take advantage of napi too.

cinderblock:
Why does `poller.h` still [`#include <nan.h>`](https://github.com/GazHank/node-serialport/blob/napi/packages/bindings/src/poller.h#L4) @GazHank?

cinderblock:
@reconbot, slightly off topic, have you considered adding coverage tests for the C/C++ sources?

GazHank:
@reconbot no worries on the timing, I feel you on the "other commitments"; I hope to be able to get some quality time on this again in the next few weeks, if I can juggle those other pesky priorities :-)

@cinderblock, et al

I'll try to address the questions and discussions in the chain, but am away from my dev machine at the moment, so appologies if some of the items will need follow up:
> What is needed to get this over the edge:

- I've tried to reflect the status / progress in the task list within the first comment and update this as I make further changes;
- In addition to the items listed, we will of course need to test on each platform (while I'm able to perform testing in Windows and Linux, I will need help to test on Mac, and with a wider selection of hardware)

> You'll only need the napi package in the bindings package, you can leave the others alone.
- Nice spot

> What's the error handling?
- I believe when I was reviewing the N-API docs the handling of errors seemed simpler and easier than NAN, I'd have to look up the specifics, but should also flag that at present I've avoided any such refactoring, and soley focused on a like for like migration where possible

> Why does poller.h still #include <nan.h>
- Yes, it's because it's a WIP, and the poller files on my local machine have only been processed via the automated coversion tool so far, so will not compile or run for a few reasons
- If others are able to help review and refine the code then I'm happy to commit those WIP files, but had planned to hold them until I had time to manually refactor the files

> So the poller is funny. I don't think that NAPI will give you anything you need to do the actual fs polling. However I at one time got assurances from the libuv team that they are ABI stable and will not change libuv's abi until their next major version. So we probably don't have to worry about until LIBUV 2.0 and just replace the nan stuff.
- I will have to take a look at some of the draft changes on my Linux machine, but from memory I think you are right.
- The way that I have tried tweaking (rather than converting) the logic within the Win10 read and write methods, was intended to be a bit of a trial for the approach I was considering for updating the poller.  If I can get those working then I think I can replicate some of the changes on that side,  Although the size of the change on the poller side is a fair bit bigger, so there may be additional changes needed that I haven't yet reviewed.

My main interest in getting N-API experts to help with the review is to make sure I'm not reinventing the wheel on any of this stuff :-D

Please feel free to query, comment and suggest changes, they are very welcome

cinderblock:
@reconbot Are the GitHub Actions running on hardware that has a serial port to test against?

Assuming yes, or even if not, the tests that are run here seem to run against all the target platforms.

If [`fail-fast`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) were set to `false`, broken platforms wouldn't prevent testing of non-broken platforms.

packages/bindings/src/darwin_list.cpp
```diff
@@ -321,20 +324,23 @@ void EIO_List(uv_work_t* req) {
...
 
-  v8::Local<v8::Value> argv[2];
+  std::vector<napi_value> args;
+  args.reserve(2);
   if (data->errorString[0]) {
-    argv[0] = v8::Exception::Error(Nan::New<v8::String>(data->errorString).ToLocalChecked());
-    argv[1] = Nan::Undefined();
+    args.push_back(Napi::String::New(env, data->errorString));
+    napi_value undefined;
+    status = napi_get_undefined(env, &undefined);
```
KevinEady, line 337: another unchecked call to a `napi_*` method. If you construct a `Napi::Env`, you can use `Napi::Env().Undefined()`

packages/bindings/src/darwin_list.cpp
```diff
@@ -344,16 +350,18 @@ void EIO_AfterList(uv_work_t* req) {
       setIfNotEmpty(item, "vendorId", (*it)->vendorId.c_str());
       setIfNotEmpty(item, "productId", (*it)->productId.c_str());
 
-      Nan::Set(results, i, item);
+      (results).Set(i, item);
     }
-    argv[0] = Nan::Null();
-    argv[1] = results;
+    napi_value null;
+    status = napi_get_null(env, &null);
```
KevinEady, line 356: Unchecked `napi_*`. Use `Napi::Env().Null()`

packages/bindings/src/darwin_list.cpp
```diff
@@ -344,16 +350,18 @@ void EIO_AfterList(uv_work_t* req) {
...
   }
-  data->callback.Call(2, argv, data);
+  data->callback.Call(args);
 
   for (std::list<ListResultItem*>::iterator it = data->results.begin(); it != data->results.end(); ++it) {
     delete *it;
   }
-  delete data;
-  delete req;
+  napi_delete_async_work(env, data->work);
```
KevinEady, line 365: Unchecked `napi_*` call, but migrating to `AsyncWorker` would alleviate this.

KevinEady (commented):
Hello from the Node-API team! This is a great start to a migration. Please see my below comments. The main concern I see is unchecked used of `napi_*` calls: when calling into the engine, you need to validate the status of `napi_ok` before making another call. The node-addon-api library does this under the hood, so if you can use the library instead of the the underlying `napi_*` calls, then you wouldn't need to add those checks yourself. This would mean using an `[AsyncWorker](https://github.com/nodejs/node-addon-api/blob/main/doc/async_worker.md)` instead of `napi_async_work`. Let us know if you have any questions!

GazHank:
Thanks @KevinEady that's super helpful, I'll try to action those changes (I think the napi_ change will be in quite a few places, so will look to update those all at once if I can).

I'm still to post the migrated poller.c and poller.h as following the use of the migration tool they still need a lot of work.  Would you be able to take a quick look at those file and let me know if there are any previous migrations of prtojects using similar functionality?

.generators/package.json
GazHank: Thanks for spotting this.

It was an artifact of the automated migration tool, so nice to tidy this up

NickNaso (commented):
Hi everybody,
this is a good start for the migration to Node-API. like suggested by @KevinEady it could be a good idea to use `Napi::AsyncWorker` instead of the C API. 
For example the following structure:
```cpp
struct VoidBaton {//: public Napi::AsyncResource {
  VoidBaton() : // AsyncResource("node-serialport:VoidBaton"), 
  errorString() {}
  int fd = 0;
  Napi::FunctionReference callback;
  napi_async_work work;
  char errorString[ERROR_STRING_SIZE];
};
``` 
could became an `Napi::AsyncWorker`:
```cpp
struct VoidBaton : public Napi::AsyncWorker {
  VoidBaton(Napi::Function& callback, int fd) : Napi::AsyncWorker(callback, "your resource name"), fd(fd)
  errorString() {}
  int fd = 0;
  char errorString[ERROR_STRING_SIZE];
};
```
Thrn you should to implement the `Napi::AsyncWorker::Execute` method to do the task you need and handle the result ny mtehods `Napi::AyncWorker::OnOK` and `Napi::AsyncWorker::OnError`.

packages/bindings/src/serialport.cpp
```diff
@@ -459,28 +508,29 @@ SerialPortStopBits NAN_INLINE(ToStopBitEnum(double stopBits)) {
...
-  Nan::SetMethod(target, "set", Set);
-  Nan::SetMethod(target, "get", Get);
-  Nan::SetMethod(target, "getBaudRate", GetBaudRate);
-  Nan::SetMethod(target, "open", Open);
-  Nan::SetMethod(target, "update", Update);
-  Nan::SetMethod(target, "close", Close);
-  Nan::SetMethod(target, "flush", Flush);
-  Nan::SetMethod(target, "drain", Drain);
+Napi::Object init(Napi::Env env, Napi::Object exports) {
+  Napi::HandleScope scope(env);
```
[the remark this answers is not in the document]
GazHank: Thanks Nick, am I correct that you mean to remove the HandleScope line?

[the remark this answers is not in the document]
NickNaso: Yes it's correct!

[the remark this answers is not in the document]
GazHank: Thanks Nick, I've bundled this into https://github.com/serialport/node-serialport/pull/2305/commits/1277ba660adca6ba86da37ccbad0608a55fc842f

packages/bindings/binding.gyp
```diff
@@ -4,9 +4,8 @@
     'sources': [
       'src/serialport.cpp'
     ],
-    'include_dirs': [
-      '<!(node -e "require(\'nan\')")'
-    ],
+    'include_dirs': ["<!@(node -p \"require('node-addon-api').include\")"],
```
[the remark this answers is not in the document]
GazHank: hi @NickNaso, I'm getting compile errors if I switch to include_dir per https://github.com/serialport/node-serialport/pull/2305/commits/eec4da98ac52b432435249cc87533059ad32fad3

With that change I recieve errors both when compiling the overall package, and when building from source as part of another project.  I assume this isn't an issue with my dev environment as reverting this change makes everything work again

GazHank:
Thank you all for the comments and feedback, it is very greatly appreciated.  I've incorporated most of the feedback so far, and hope to get some time tomorrow to look into the switch to Napi::AsyncWorker as this seems to be a potentially larger and more significant change.

packages/bindings/binding.gyp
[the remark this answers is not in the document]
GazHank: It looks like the @ just needed to be removed when switching to include_dir

packages/bindings/src/serialport_unix.cpp
```diff
@@ -76,7 +76,7 @@ int ToDataBitsConstant(int dataBits) {
   return -1;
 }
 
-void EIO_Open(napi_env env, void* req) {
+void EIO_Open(void* req) {
```
GazHank, line 79: env was never used, and I believe it should not be used within the Execute method of the AsyncWorker

packages/bindings/src/serialport_win.cpp
```diff
@@ -56,7 +56,7 @@ void AsyncCloseCallback(uv_handle_t* handle) {
   delete async;
 }
 
-void EIO_Open(napi_env env, void* req) {
+void EIO_Open(void* req) {
```
GazHank, line 59: env was never used, and I believe it should not be used within the Execute method of the AsyncWorker

packages/bindings/src/serialport.h
```diff
@@ -88,6 +85,21 @@ struct OpenBaton {//: public Napi::AsyncResource {
   uint8_t vmin = 0;
   uint8_t vtime = 0;
 #endif
+  void Execute() override {
+    EIO_Open(this);
+  }
+
+  void OnError(Napi::Error const &error) override {
```
GazHank, line 92: Logic from The EIO_AfterOpen method

packages/bindings/src/serialport.cpp
```diff
@@ -66,41 +66,16 @@ Napi::Value Open(const Napi::CallbackInfo& info) {
...
 
-  napi_value resource_name;
-  napi_create_string_utf8(env, "Open", NAPI_AUTO_LENGTH, &resource_name);
-  napi_create_async_work(env, NULL, resource_name, EIO_Open, EIO_AfterOpen, baton, &baton->work);
-  napi_queue_async_work(env, baton->work);
+  baton->Queue();
   return env.Undefined();
 }
 
-void EIO_AfterOpen(napi_env n_env, napi_status status, void* req) {
```
GazHank: Logic transfererd to the OnError and OnOK within the struct

GazHank:
Hi @KevinEady & @NickNaso 

I've tried to convert OpenBaton to use AsyncWorker per your suggestions, so would be very greatful if you could take a look at those changes; I think it's probably sensible to include any further revisions to that before I roll out similar changes to the other Structs and Methods.

The changes can be found in  https://github.com/serialport/node-serialport/pull/2305/commits/d042c04f238427bb3b3547f00c8dfeec1ffd5e61

I've added some comments to the commit, but to summarise:
Updated struct definition
- Switch AsyncResource to Napi::AsyncWorker
- Remove callback and work (provided by AsyncWorker instead)
- Transfer EIO_AfterOpen logic (from serialport.cpp) into the OnError and OnOK within the struct (serialport.h)
- Override Execute to call EIO_Open (since there is discrete logic for Windows and Linux, I've tried not to touch these, except removing the unused env within the method call)

The changes to the Open method are:
- Replace napi_queue_async_work with AsyncWorker->Queue()
- Revise how the callback function is passed

This change seems to work based on initial testing, but I've not tried to invoke error conditions, so am not sure if the override of OnError is correct, or the best way to handle things...

GazHank:
> This method will be called only if you set the error using the method `Napi::AsyncWorker::SetError` otherwise the `Napi::AsyncWorker::OnOK` method will be called.
> See: https://github.com/nodejs/node-addon-api/blob/main/doc/async_worker.md#seterror and https://github.com/nodejs/node-addon-api/blob/main/doc/async_worker.md#execute
> One idea is to call the `Napi::AsyncWorker::SetError` at the end of the `Napi::AsyncWorker::Execute` method like repoted below:
> 
> ```c++
>   void Execute() override {
>     EIO_Open(this);
>    SetError(std::string(errorString));
>   }
> ```

Thanks @NickNaso I assume this is preferable to simply including this logic in the OnOK call?

NickNaso:
> > This method will be called only if you set the error using the method `Napi::AsyncWorker::SetError` otherwise the `Napi::AsyncWorker::OnOK` method will be called.
> > See: https://github.com/nodejs/node-addon-api/blob/main/doc/async_worker.md#seterror and https://github.com/nodejs/node-addon-api/blob/main/doc/async_worker.md#execute
> > One idea is to call the `Napi::AsyncWorker::SetError` at the end of the `Napi::AsyncWorker::Execute` method like repoted below:
> > ```c++
> >   void Execute() override {
> >     EIO_Open(this);
> >    SetError(std::string(errorString));
> >   }
> > ```
> 
> Thanks @NickNaso I assume this is preferable to simply including this logic in the OnOK call?

Yes  you are right.

GazHank:
Thanks for all of the feedback!

I've migrated all of the common functions to avoid native napi_ calls.  The EIO_After functions have been converted to OnError and OnOK within the struct defintions.

Instead of invoking the EIO functions within the Execute() function of the struct, I've directly converted the EIO functions into overrides of the Execute() functions.  I think this is cleaner and simpler, but would be keen to get feedback on this (I'd be happy to revert this to the approach previously taken with setBaton if that is preferable).

I seem to have introduced a defect into the Get function as the GetBaton OnOK doesnt seem to be returning the results object.  I'll try to get this fixed before converting the List and other platform specific logic (i.e. read, write and poller).

ps - I was wondering if Napi::AsyncProgressWorker might be a good fit for the read and write functions for windows, but havent had chance to look into that much as yet.

GazHank:
I've made an initial change to convert the Windows specific read and write functions to Napi::AsyncWorker.  I wonder if Napi::AsyncProgressWorker or Napi::AsyncProgressQueueWorker would be more appropriate.

It also looks like the event handling needs to be updated as `SerialPort.on('close', function() {console.log("port closed");})` currently doesn't work at all.

GazHank:
Some of the common logic requried a few tweaks to build on Linux, I'm not sure if this is due to stricter build rules, or platform differences, but it suggests that we will need someone to test this on macOS (once the poller updates are implemented).

At the moment it appears that the 2 main items to work on are poller for Mac and Linux and the event handling for Windows.  I hope to get some time to look at those in the next few days, but in the meantime please review/critique and suggest updates, the support from the Node-API team and community has been a great help!

packages/bindings/src/serialport.h
```diff
@@ -3,49 +3,48 @@
 
 // Workaround for electron 11 abi issue https://github.com/serialport/node-serialport/issues/2191
 #include <node_version.h>
-#if CHECK_NODE_MODULE_VERSION && NODE_MODULE_VERSION == 85
+#if CHECK_NODE_API_MODULE_VERSION && NODE_API_MODULE_VERSION == 85
```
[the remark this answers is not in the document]
GazHank: Thanks @gabrielschulhof I've added this to the TODO list for now, while focussing on broken functionality, but have linked your comment so as to not lose this change.

GazHank:
I'm not sure if it is the Linux changes or implementing close and other simple methods via inheritance of the voidBaton, but the close event seems to work again :-D

Windows should now be ready for some testing for anyone who feels adventurous.

Now it's time for me to look into the poller :-(  wish me luck

GazHank:
I've proposed a few changes to the poller.  Some of this is actually just reverting changes introduced by the migration script, so I'm not certain if it is moving us forward or backwards... however it will now compile on Linux, but the poller is still a bit broken.  I've flagged a couple of obvious issues with `// TODO FIX` as I need to look into those a little more.

So far I've simply maintained the ObjectWrap and use of  uv.h; as I don't know if there AsyncProgressWorker or something similar would be a better fit.

Sorry for committing a change which still leaves the poller non functional, but I'm not sure how much time I will get to look at this over the next week (back to the day job I'm afraid); so I wanted to share an update in case others are able to take a look in my absence.

packages/bindings/binding.gyp
```diff
@@ -4,9 +4,9 @@
     'sources': [
       'src/serialport.cpp'
     ],
-    'include_dirs': [
-      '<!(node -e "require(\'nan\')")'
-    ],
+    'include_dirs': ["<!(node -p \"require('node-addon-api').include_dir\")"],
+    'dependencies': ["<!(node -p \"require('node-addon-api').gyp\")"],
+    "defines": ["NAPI_DISABLE_CPP_EXCEPTIONS"],
```
cinderblock, line 9: 1. Why not use C++ exceptions?
2. From `node-addon-api` [docs](https://github.com/nodejs/node-addon-api/blob/main/doc/setup.md):
   > If you decide to use `node-addon-api` without C++ exceptions enabled, please consider enabling node-addon-api safe API type guards to ensure the proper exception handling pattern:
   > ```gyp
   > 'defines': [ 'NODE_ADDON_API_ENABLE_MAYBE' ],
   > ```

packages/bindings/binding.gyp
```diff
@@ -4,9 +4,9 @@
     'sources': [
       'src/serialport.cpp'
     ],
-    'include_dirs': [
-      '<!(node -e "require(\'nan\')")'
-    ],
+    'include_dirs': ["<!(node -p \"require('node-addon-api').include_dir\")"],
```
cinderblock, line 7: I've always seen this setting as:

```gyp
'include_dirs': ["<!@(node -p \"require('node-addon-api').include_dir\")"],
```

Specifically, the `@` seems to be missing.

Note, `dependencies` generally doesn't use the `@`.

GazHank:
Some progress on the Linux front.

- I've implemented some fixes to the Open/setup methods, the error handling for OpenBaton has me scratching my head as it doesnt seem to throw errors properly even though the OnError method is triggered (and the logic seems identical to other bits of code that do work).
- Poller has some fixes.
- NODE_ADDON_API_ENABLE_MAYBE added (Per suggestion from @cinderblock)  at the moment NAPI_DISABLE_CPP_EXCEPTIONS is still needed as without this the compiler just complains about napi.h and napi-inl.h (they are probably down to errors or warnings in the code for this package, but the compiler errors are not helpful to trace the problem).
- I've also proposed a couple of changes to address compiler warnings - one for the inline function syntax, and one for truncation of the error string.  These aren't needed (they just got a little annoying when I was recompiling frequently).  Happy to revert or refactor those changes if desired.

I'll try to find some more time over the weekend to iron out some more of the problems (I've not check that these changes don't cause regressions on the Windows side as yet, so also plan to check that)

GazHank:
So it looks like the onError logic was the problem with the error handling.  Windows is still functioning, and now 'npm test' is returning the same result as the latest release (maily successful, but some of the test just don't work on my local hardware even with the production release).

Back to Linux next...

GazHank:
It's been some time, so I thought it useful to give a brief update.

On the positive side: windows build seems pretty good for all of my tests; in addition I'm able to enable CPP Exceptions, and upgrade to electron 14 and/or windows 11 without any problems.

On the negative side: I've chased segfaults around on Linux for a little too long.  My debug setup on Linux isn't too good as it's not a platform that I use.  I can't be sure if the issues are due to an error in the way I'm trying to use UV or N-API.  For example, I can avoid all of the segfault issues if I don't try to destroy the poller, and just wait for the garbage collection when the program terminates; but that doesn't seem sensible.  Also there is some existing logic to avoid segfaults when UV stop is called, but they don't seem to protect all scenarios (I can only assume these are secondary as they work fine with the current build).

If anyone is able to take a look at the poller code and make any suggestions that would be greatly appreciated.  I've made lots of tweaks and changes locally, but have avoided committing any changes until I know they are improving the situation.

Many thanks

Gareth

GazHank:
Ok, so I finally found a little time to look at the poller :-)

This should finally be in a suitable state to start testing.  So far I've run local tests on Windows and Linux and am getting consistent positive results for my limited test cases.  I would be very grateful if others could start to take a look, especially on Mac.

packages/bindings/binding.gyp
GazHank: Thanks @cinderblock, I've enabled the CPP exceptions in the latest set of changes

packages/bindings/binding.gyp
GazHank: This seems to have been revised as part of https://github.com/nodejs/node-addon-api/commit/f27623ff611645cd7d927f0de81158a93e960542#diff-55ad3a91265d731555313fdee9af5e109244b0e825c4aff0b7cce3c78b07be97

GazHank:
> @GazHank I am waiting for this PR to be merged because I need it for an electron 15 project, I really appreciate your efforts here!
> 
> I am not really a NAPI developer so I am not able to help much on that front, but I just wanted to let you know that I was able to use your branch on MacOS with node 14 and node 16. All tests are running successfully and I was able to list my devices, connect to my barcode reader and received valid data. Unfortunately I was not able to test it in electron since I am not really sure how to link the project properly from the mono repo.
> 
> If you want me to test something on MacOS I am happy to help!

That's excellent thanks @thegnuu!

If you would like to try testing this in your electron project (or included in any other project really) then I think the easiest way to to download the PR locally, and then add that local copy as a resolution within your project's package.json.

For example:

```
    "dependencies": {
        "serialport": "^9.2.0"
    },
    "resolutions": {
        "@serialport/bindings": "file:../node-serialport/packages/bindings"
    }
```

You just need to make sure the directory matches your install location, and then run yarn install etc

If you find any issues then please let me know

PrestonGiorgianni:
> > @GazHank I am waiting for this PR to be merged because I need it for an electron 15 project, I really appreciate your efforts here!
> > I am not really a NAPI developer so I am not able to help much on that front, but I just wanted to let you know that I was able to use your branch on MacOS with node 14 and node 16. All tests are running successfully and I was able to list my devices, connect to my barcode reader and received valid data. Unfortunately I was not able to test it in electron since I am not really sure how to link the project properly from the mono repo.
> > If you want me to test something on MacOS I am happy to help!
> 
> That's excellent thanks @thegnuu!
> 
> If you would like to try testing this in your electron project (or included in any other project really) then I think the easiest way to to download the PR locally, and then add that local copy as a resolution within your project's package.json.
> 
> For example:
> 
> ```
>     "dependencies": {
>         "serialport": "^9.2.0"
>     },
>     "resolutions": {
>         "@serialport/bindings": "file:../node-serialport/packages/bindings"
>     }
> ```
> 
> You just need to make sure the directory matches your install location, and then run yarn install etc
> 
> If you find any issues then please let me know

using [npm link](https://docs.npmjs.com/cli/v7/commands/npm-link) is also an option.

```bash
cd ../node-serialport/packages/bindings
npm link
cd /my/local/project
npm link @serialport/bindings
```

thegnuu:
Thank for the tips! I was aware of those two methods, but it was not working as expected before. I guess my issue was that I did not update the version number of `@serialport/bindings` which caused `electron-builder` to download the prebuilt binaries for the version set in `package.json`.

After changing the version to `10.0.0` electron-builder was able to rebuild the local binaries and as far as I can tell it works without any issues. Currently I have only the `list` operation implemented in electron (it's a new feature in our software), but I will try to connect to a device and read some data tomorrow and let you know. Unfortunately I don't have a device to send data to and test this part.

cinderblock:
@thegnuu I don't mean to discourage you because more testing here is helpful but just wanted to make sure you knew...

The traditional way of making `serialport` work with electron is with the [`electron-rebuild`](https://www.npmjs.com/package/electron-rebuild) package. If you need this support today, does that work for you?

thegnuu:
> @thegnuu I don't mean to discourage you because more testing here is helpful but just wanted to make sure you knew...
> 
> The traditional way of making `serialport` work with electron is with the [`electron-rebuild`](https://www.npmjs.com/package/electron-rebuild) package. If you need this support today, does that work for you?

@cinderblock you are of course right and my comment was not detailed enough. We are using `electron-builder` to package, sign and notarize our application, so we use the (not really well documented) command `electron-builder install-app-deps` to recompile the native modules. This does basically the same as `electron-rebuild` and since it is already bundled with `electron-builder` we just use this one. Should have pointed out the command which is used not the package alone....

thegnuu:
@GazHank I was now able to test the following with electron 15.1.0 (node 16.5.0) on MacOS Big Sur 11.4 successfully:

- List all devices
- Connect to my datalogic barcode reader via usb serial
- Getting data when I read a code
- Closing the connection

As mentioned before those cases work as well without any issues with node 14 and 16.

I hope this helps, if I should test anything else please let me know :)

GazHank:
I've finally got an arduino device, so that I can run the full set of automated test scripts.  I'll run the tests on Window and Linux, but wont be able to test on Mac.

For anyone with a Mac and Arduino device, the arduino sketch required to enable testing can be found at [arduinoEcho](https://github.com/serialport/node-serialport/blob/ac2a64ba4bf35ef77b0bfd4b307212e158be6a6b/test/arduinoTest/arduinoEcho/arduinoEcho.ino)  

So far on Windows I have 4 issues to resolve:

```
  1) bindings/win32
       instance method
         #read
           cancels the read when the port is closed:
     Error: Should have rejected
      at global.shouldReject (test\initializers\assert-reject.js:8:9)

  2) win32 SerialPort Integration Tests
       opening and closing
         errors if closing during a write:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\Git\napi\node-serialport\packages\serialport\lib\serialport.test.js)
      at listOnTimeout (node:internal/timers:557:17)
      at processTimers (node:internal/timers:500:7)

  3) win32 SerialPort Integration Tests
       #flush
         discards any received data:
     Error: got a readable event after flushing the port
      at SerialPort.<anonymous> (packages\serialport\lib\serialport.test.js:219:22)
      at SerialPort.emit (node:events:390:28)
      at emitReadable_ (node:internal/streams/readable:578:12)
      at onEofChunk (node:internal/streams/readable:556:5)
      at readableAddChunk (node:internal/streams/readable:269:5)
      at SerialPort.Readable.push (node:internal/streams/readable:228:10)
      at C:\Git\napi\node-serialport\packages\stream\lib\index.js:99:194

  4) win32 SerialPort Integration Tests
       #flush
         discards any received data:
     Error: done() called multiple times in test <win32 SerialPort Integration Tests #flush discards any received data> of file C:\Git\napi\node-serialport\packages\serialport\lib\serialport.test.js
      at C:\Git\napi\node-serialport\packages\stream\lib\index.js:111:1107

```

thegnuu:
@GazHank I found a spare arduino uno laying around somewhere in a drawer, quickly flashed the script and I was able to run the tests without any issues!

```
  323 passing (14s)
  3 pending
```

The three tests that are pending are intentionally disabled on MacOS I guess:

```
- Feature "baudrate.25000_check" is disabled in "darwin. "sets 25000 baud successfully"
- Feature "baudrate.1000000_check" is disabled in "darwin. "sets 1000000 baud successfully"
- Feature "baudrate.250000_check" is disabled in "darwin. "sets 250000 baud successfully"
```

AntonioDeCasper:
@thegnuu thanks for code example. Lately i try to import SerialPort and use it right in the my preload script:

```js
const mainWindow = new BrowserWindow({
    title: '${app.name} v. ${app.getVersion()}',
    webPreferences: {
      nodeIntegration: true,
      contextIsolation: false,
      preload: join(__dirname, "preload.js"),
    },
  });
```

but it not work. 
Today i try use it right in the  main electron file and got a list of ports available:

```js
webContents.on("did-finish-load", () => {
    SerialPort.list().then((portList) => {
      console.log("portList: ", portList);
    });
});
```
Thanks for helping. I will test this PR too. @GazHank thank you for your work, its awesome!

GazHank:
I've finally fixed the issues on Windows with the Read and Write operations. All 322 tests are passing now on Win11.

On Linux I'm getting the exact same result with this version and the current production version (a few timeouts, but absolutely identical to current production release)

So at the moment we seem to be at  💯% success on the automated testing, its time for real world testing to see if we can break it.  Please do your worst, and report back any issues you find!

Additionally, I'm considering removing the current hack/patch that is used to workaround an [issue on Electron 11](https://github.com/serialport/node-serialport/issues/2191).  Since [electron 11 is no longer supported](https://www.electronjs.org/docs/tutorial/support#end-of-life), this migration to N-API seems like a sensible time to drop this workaround.  If anyone has stong opinions about this then let me know and I'd be happy to reconsider

packages/bindings/src/poller.cpp
```diff
@@ -31,112 +48,119 @@ void Poller::onClose(uv_handle_t* poll_handle) {
...
-NAN_METHOD(Poller::destroy) {
-  Poller* obj = Nan::ObjectWrap::Unwrap<Poller>(info.Holder());
-  obj->persistent().Reset();
-  delete obj;
+Napi::Value Poller::destroy(const Napi::CallbackInfo& info) {
+  Napi::Env env = info.Env();
+  Napi::HandleScope scope(env);
+  Poller* obj = this;
+  // TODO Fix destruction Segfault
+  obj->Reset();
```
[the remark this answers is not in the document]
GazHank: Unfortunately the automated testing doesn't generate the seg fault; I've tried recreating the scenario within mocha, but at the moment I can only get it to trigger within a more complex piece of code.  It seems to occur upon opening and closing a connection twice, but that test case still works within the mocha tests :-(

I'll continue to see if I can reproduce the error in a smaller set of code or automated test case

clostao:
Hi, is the implementation of migration to NAPI completed? In that case, how could I use it as an npm package?

GazHank:
> Hi, is the implementation of migration to NAPI completed? In that case, how could I use it as an npm package?

Hi @clostao at the moment the migration isn't ready for production deployments, but it is ready for testing.  I'm hoping to get time to publish this as a beta v10 release soon (with the associated prebuilt binaries); but in the meantime you should be able to build based on this PR provided you have the usual build tools installed on your machine.

GazHank:
Thanks @reconbot 

sorry I've been away for a while, hopefully work and travel will calm down a bit as we move towards the end of the year

brandonros:
> @GazHank want to rebase and merge? We can cut a beta release from here.

could you cut that beta release?

thegnuu:
It would be great to have that beta release soon, it would allow me to implement the changes in our app and make some tests out in the field :)

Eugeny:
Yep, I'll be glad to push it into the Tabby nightly and get feedback.

GazHank:
I'd like to extend a big thank you to everyone who has supported this; it's been wonderful to see the level of community engagement and contribution!

Eugeny:
Parity options do not work on windows, because the temporary string in `ToParityEnum` gets destroyed before the comparison, leading to an invalid pointer in `str`. 

Fix:

```diff
diff --git a/node_modules/@serialport/bindings/src/serialport.cpp b/node_modules/@serialport/bindings/src/serialport.cpp
index c48e150..00a5f5a 100644
--- a/node_modules/@serialport/bindings/src/serialport.cpp
+++ b/node_modules/@serialport/bindings/src/serialport.cpp
@@ -269,7 +269,8 @@ Napi::Value Drain(const Napi::CallbackInfo& info) {
 }

 inline SerialPortParity ToParityEnum(const Napi::String& napistr) {
-  const char* str = napistr.Utf8Value().c_str();
+  auto tmp = napistr.Utf8Value();
+  const char* str = tmp.c_str();
   size_t count = strlen(str);
   SerialPortParity parity = SERIALPORT_PARITY_NONE;
   if (!strncasecmp(str, "none", count)) {

```

brandonros:
Has anybody been able to build this with Electron on Windows?

```
PS C:\Users\Brandon\Documents\GitHub\project\foo> npm install

> foo@1.2.0 postinstall
> electron-builder install-app-deps

  • electron-builder  version=22.14.5
  • loaded configuration  file=package.json ("build" field)
  • rebuilding native dependencies  dependencies=@serialport/bindings-cpp@10.2.1, ffi-napi@4.0.3, ref-napi@3.0.3, usb@1.8.0 platform=win32 arch=ia32
  ⨯ cannot execute  cause=exit status 1
                    errorOut=npm ERR! code 1
    npm ERR! path C:\Users\Brandon\Documents\GitHub\project\foo\node_modules\@serialport\bindings-cpp
    npm ERR! command failed
    npm ERR! command C:\Windows\system32\cmd.exe /d /s /c node-gyp-build
    npm ERR! Building the projects in this solution one at a time. To enable parallel build, please add the "-m" switch.
    npm ERR!   serialport.cpp
    npm ERR! C:\Users\Brandon\.electron-gyp\16.0.7\include\node\node_api.h(12,10): fatal error C1083: Cannot open include file: 'js_native_api.h': No such file or directory [C:\Users\Brandon\Documents\GitHub\project\foo\node_modules\@serialport\bindings-cpp\build\bindings.vcxproj]
```

PR #2317, opened 2021-09-16
Quick low latency fix

GazHank:
Avoided throwing errors when low latency fails.

Disabled errors:
- Low latency failure during Get function - simply returns 'low latency = false'
- Low latency failure during Set function when the user did not specify low latency as true

Errors only occur if we are sure the user requested to set low latency and some aspect of the process fails.

This doesn't really handle the situation where someone is trying to toggle Low Latency mode on and off since only the enable request would result in an error being reported.

This may be a short lived patch for #2312 until a longer term code change is implemented.  We don't really have an existing API to adopt for this feature AFAIK, but it seems sensible to rework this into the Open and Update methods, rather than in the Set and Get methods


reconbot (approved):
lgtm

PR #2318, opened 2021-09-20
chore: allow monthly dependabot for all packages

reconbot:


GazHank:
Looks good.

I think there will be a bit of a flood of updates when this first activates.  I think that by default the bot maxes out at 5 active PRs, but am not sure how this will work with the monthly setting - will it just raise 5 each month, or will it raise extras when those are accepted...

GazHank (approved):

Issue #2259, opened 2021-06-03
SerialPort.list() crashes Node when used with brother device


rb-serin:
### SerialPort.list() crashes Node ( used on electron) when a Brother printer is attached

(Please answer all 3)

- What are you trying to do?
Listing all devices attached to the machine

- What happens?
Node crashes leaving a blank page on electron renderer process

- What should have happened?
Getting a list of devices or an exception

### Code to Reproduce the Issue

```js
const SerialPort = require('serialport')
SerialPort.list()

```

### Versions, Operating System and Hardware

- SerialPort@ 7.x and 9.x 
- Node.js v14.15.4
- Windows? Linux? Mac? Windows
- Hardware and chipset? (Prolific/FTDI/Other) Intel core i5 with Brother printer attached

This is the same problem as #1518
See also https://www.gitmemory.com/issue/serialport/node-serialport/2155/771687008

GazHank:
Hi @rb-serin

Would you be able to confirm the specific version(s) of electron and serial port that you are using?

Also, could you confirm if you are facing the issue of the device not being assigned a COM port per the prior issue you referenced?

If the COM port isn't being assigned by windows you might want to manually configure the COM port within device manager (there are some instructions here: https://www.virtual-serial-port.org/articles/com-port-on-windows-10/#assign ).  Although even if that does resolve your issue it sounds like the error handling/ fault tolerance could be improved.

GazHank:
This might be related to https://github.com/serialport/node-serialport/issues/2155 as both seem to relate to unexpected behaviour; in both cases I think we need more info to be able to investigate and debug

nyholmjan:
I can reliably reproduce this issue. The key part is to query for ports while windows is installing usb to serial drivers. It can be reproduced simply by removing the usb to serial drivers of the device from device manager and then unplugging the device and plugging it in again, the crash will happen when serialport.list() is called while the drivers are being reinstalled.

## Versions, Operating System and Hardware
- serialport: 9.2.0
- electron: 13.2.1
- os: windows 10
- hardware: u-blox EVK-M101

## Code to reproduce
```
const serialport = require("serialport");
window.setInterval(() => {
    serialport.list().then(ports => {
        console.log(ports)
}, 1000);
```

GazHank:
[...]

Issue #2155, opened 2020-10-16
Segfaults during `list()` that crashes Node when encountering certain types of serial ports


petertorelli:
# Summary of Problem

(Please answer all 3)

- What are you trying to do? Query the serial ports.
- What happens? I get a segmentation fault when I plug in a particular device.
- What should have happened? I should have seen the new device added.

### Code to Reproduce the Issue

```js
const serialport = require("serialport");
serialport.list().then(currentPorts => {
	console.log(currentPorts);
});
```

### Versions, Operating System and Hardware

- SerialPort@? 9.0.2
- Node.js v? 12.18.0
- Windows? Linux? Mac? Windows 10
- Hardware and chipset? Dialog DA16200 debug port

The DA16200 appears to create a SerialPort device that windows cannot assign to a COM port. Unfortunately, the `node-serialport` module segfaults with no indication of an error when it encounters this device: I can't even `catch()` it and recover, the process just ends. Even when I run with `DEBUG=*` I see no other messages. Windows Command Prompt prints nothing but Cygwin offers a clue:

```
Peter@CAPYBARA /cygdrive/c/Users/Peter/serialtmp
$ DEBUG=* node test-serial.js
  serialport/bindings loading WindowsBinding +0ms
  serialport/stream .list +0ms
Segmentation fault
```

Here is a picture of the orphan Serial Port appearing in Device Manager:

![image](https://user-images.githubusercontent.com/8249735/96287867-63932c80-0f97-11eb-8f7e-317d1a168c57.png)

How do I enable better error handling so that my application doesn't segfault due to serialport?

gniezen:
I can confirm that the same issue is happening for our users (at least 5) who have devices listed under "Ports (COM & LPT)" without a COM port number. In one specific instance, the device without COM port number is a Brother HL-L2350DW printer.

Serialport 9.0.2
Node 12.13.0
Windows 10

Ours is an Electron app, so when the segfault occurs the entire window just goes blank, without any error messages. It took us forever to figure out what was going on, so thank you so much for reporting this @petertorelli !

Any workarounds for this one, @reconbot ?

petertorelli:
@gniezen FYI: This is also an ElectronJS application, but it also happens under a plain NodeJS process as well.

reconbot:
[...]

Issue #1830, opened 2019-04-04
SerialPort 7.1.4 crashing on Win64 when calling list()


caffeinatedbits:
### Summary of Problem
My application gets a list of serial ports, and then makes a connection to a device to transfer data. At the point in which serialport .list() is called, the application is crashing with Crash reason:  EXCEPTION_ACCESS_VIOLATION_READ in bindings.node (dumps attached). I've got this app installed on ~500 end user systems, but the error is only occurring on one, a laptop with Win10 Pro x64.

(Please answer all 3)
[dump1.txt](https://github.com/node-serialport/node-serialport/files/3045496/dump1.txt)
[dump2.txt](https://github.com/node-serialport/node-serialport/files/3045497/dump2.txt)
[dump3.txt](https://github.com/node-serialport/node-serialport/files/3045498/dump3.txt)

- What are you trying to do?
List serial ports

- What happens?
Application crashes (dump generated)

- What should have happened?
Serial ports should be listed

### Code to Reproduce the Issue

```js
// Code
const SerialPort = nw.require('serialport');
SerialPort.list().then( (ports) => { ...code... }, (error) => { ...code... });
```

### Versions, Operating System and Hardware

- SerialPort@7.14
- Node.js v? - 11.10.0 / NWJS 0.36.3
- Windows? Linux? Mac? - Windows 10 x64
- Hardware and chipset? (Prolific/FTDI/Other) - Unknown USB-to-RS232 Serial Adapter

caffeinatedbits:
Update: I've now discovered a total of 3 systems with this same issue, and they all have a Brother multifunction printer attached by USB. When the printer cable is disconnected, my app runs flawlessly. Why would a Brother printer connected via USB cause issues with serialport?

reconbot:
It would be the brother print drivers that don't like to be querried for device info. We use USB information to get more info on each serialport.

reconbot:
You can try the [`usb`](https://www.npmjs.com/package/usb) npm package to see if you can find your device in a way that wont mess up the printers. I'd also file a bug with Brother.

PR #2325, opened 2021-09-29
Fixed EIO_List to prevent crash when a device has no COM port

caffeinatedbits:
Fixed EIO_List to prevent crash when a device has no COM port, see:
https://github.com/serialport/node-serialport/issues/1830
https://github.com/serialport/node-serialport/issues/2259
https://github.com/serialport/node-serialport/issues/2155


GazHank:
While I don't have a brother device that exhibits this issue, I can confirm that this doesn't seem to cause regressions for normal windows usage :-)

The only other item which isn't initialised upon each loop is serialNumber... not sure if that is worth a quick change in case that is causing any other issues...

PR #2359, opened 2021-11-24
update "prebuild-install" to "^7.0.0"

zixuan945:
"prebuild-install":"^6.1.4" has a dependency 'node-abi':'^2.21.0', which does not get the electron node-abi version correctly
https://github.com/prebuild/prebuild-install/pull/161#issue-1046743812
> [electron/node-abi#113](https://github.com/electron/node-abi/issues/113) was fixed in 3.3.0 and is affecting electron 14+
> 
> With the current node-abi prebuild-install will download ABI 89 for electron 14 and 15 and then modules will fail to load as electron 14 has ABI 97 and electon 15 has ABI 98 in reality.

### DIFF
--- a/packages/bindings/package-lock.json
+++ b/packages/bindings/package-lock.json
@@ -1,1125 +1,1159 @@
 {
-	"name": "@serialport/bindings",
-	"version": "9.2.7",
-	"lockfileVersion": 2,
-	"requires": true,
-	"packages": {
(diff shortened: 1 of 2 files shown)


zixuan945:
Yes, I did set up a custom registry. I 'm not familiar with lock file
Thank you for fixing this issue.

PR #2314, opened 2021-09-06
Generate C++ coverage reports too!

cinderblock:
Coverage tests are wonderful.

This PR adds support for running coverage tests on this package's [**C++** sources](https://codecov.io/gh/serialport/node-serialport/tree/4f4d9e89698a311ab6bb6c8504c09820e3109357/packages/bindings/src) _alongside_ **JavaScript**'s!

| | [Before](https://codecov.io/gh/serialport/node-serialport/tree/c8ca2fcb789955d3f4ad9dc9a8bdd4dff1e8cb4a) | [After](https://codecov.io/gh/serialport/node-serialport/tree/4f4d9e89698a311ab6bb6c8504c09820e3109357) | &nbsp;&nbsp;&nbsp;[Change](https://app.codecov.io/gh/serialport/node-serialport/compare/2314/tree) |
| --- | ---: | ---: | ---: |
| Covered&nbsp;Lines   | 1 019 | 1 397 | +&nbsp;37% |
| Uncovered&nbsp;Lines   | 96 | 1 219 |  +&nbsp;1 170% |
| Tracked&nbsp;Lines   | 1 115 | 2 616 | +&nbsp;135% |
| Coverage | 91% | 53% | ⚠️-&nbsp;42% |
| Sunburst | [Sunburst Before](https://codecov.io/gh/serialport/node-serialport/commit/c8ca2fcb789955d3f4ad9dc9a8bdd4dff1e8cb4a/graphs/sunburst) | Update: codecov broke this image!<br/>[Sunburst After ](https://codecov.io/gh/serialport/node-serialport/commit/33bc2401ed2a9611d5642ac2372e54c53d67d6b3/graphs/sunburst)| |

### [Before](https://codecov.io/gh/serialport/node-serialport/tree/c8ca2fcb789955d3f4ad9dc9a8bdd4dff1e8cb4a/packages/bindings)

[Before C++ coverage tracking](https://codecov.io/gh/serialport/node-serialport/tree/c8ca2fcb789955d3f4ad9dc9a8bdd4dff1e8cb4a/packages/bindings)

### [After](https://codecov.io/gh/serialport/node-serialport/tree/33bc2401ed2a9611d5642ac2372e54c53d67d6b3/packages/bindings)
Update: codecov broke this image!<br/>
[After C++ coverage tracking](https://codecov.io/gh/serialport/node-serialport/tree/33bc2401ed2a9611d5642ac2372e54c53d67d6b3/packages/bindings)

### Platforms & Issues

[...]

PR #2363, opened 2021-12-04
CI cleanup

cinderblock:
Remove extraneous workflow and slightly improve outputs, as a separate PR, [as requested](https://github.com/serialport/node-serialport/pull/2314#discussion_r723714127).


package.json
```diff
@@ -9,15 +9,15 @@
...
-    "coverage": "nyc report --reporter=text-lcov > coverage.lcov && codecov",
     "format": "eslint packages test --fix",
     "generate": "plop",
     "lint": "eslint packages test && lerna run --stream lint",
     "outdated": "lerna exec --no-bail npm outdated && npm outdated",
     "postinstall": "lerna bootstrap --no-ci",
     "prebuild": "lerna run --stream prebuild",
     "publish": "lerna publish --exact",
-    "test": "nyc --reporter=html --reporter=text --reporter lcovonly mocha",
+    "coverage:publish:node": "codecov --disable=gcov --file=coverage/lcov.info",
```
reconbot, line 19: What does the new name mean?

cinderblock:
On its own, in this PR, it doesn't do anything.

Once the different OSes have more steps, as proposed in #2314, I, personally, found it a little easier to read broken up by OS. Having this change here makes the later diff more obvious.

package.json
cinderblock: IMHO, this is the main part that needs some workshopping, in conjunction with #2314. Each OS has slightly different steps for collecting coverage reports and publishing the data to codecov.

Previously, `scripts#coverage` reran the tests to generate the raw coverage data a second time, before uploading to codecov, to ensure the coverage files were all generated. I felt this was redundant with just running the `test` script first.

As for why I picked this script name, it lines up with names chosen in #2314. But, again, I'm rather open to these names changing.

Thinking about this more, codecov *should* be able to hand some of the cross-platform issues. I'll look at that to see if I can simplify the coverage steps.

reconbot: Previously it should have just been uploading the existing code coverage from the previous step.

cinderblock: Currently, the [CI action runs](https://github.com/serialport/node-serialport/blob/master/.github/workflows/test.yml#L31-L32):

```bash
npm test
npm run coverage
```

Which, respectively, just run:

```bash
nyc --reporter=html --reporter=text --reporter lcovonly mocha
nyc report --reporter=text-lcov > coverage.lcov && codecov
```

While the `nyc report --reporter=text-lcov > coverage.lcov` part generates `coverage.lcov`, `codecov` doesn't need it. If you look at the [CI logs](https://github.com/serialport/node-serialport/runs/4330567254?check_suite_focus=true), you'll see that `codecov` finds two identical coverage files:

```log
==> Scanning for reports
    + /home/runner/work/node-serialport/node-serialport/coverage/lcov.info
    + /home/runner/work/node-serialport/node-serialport/coverage.lcov
```

`coverage/lcov.info` is generated by the `nyc ... mocha` with the `lcovonly` "reporter" enabled and has the exact same information as `text-lcov`. So the `nyc report ... > coverage.lcov` is a little redundant.

Looking at this more closely, I think I may have previously missed the `report` command and just assumed it was fully rerunning the tests. This may have been incorrect.

[Looking at the docs for `report`](https://www.npmjs.com/package/nyc#combining-reports-from-multiple-runs), it seems the intent is to use `report` in more complicated testing scenarios where a few tests are run ahead of time and the final reports are generated as a final step. This could be valuable but currently `codecov` also handles merging coverage information from multiple related runs so it doesn't seem as necessary to me.

reconbot: you were right - I cleaned it up

PR #2365, opened 2021-12-12
feat: move to prebuildify

reconbot:
TODO
- missing github action's level building


thegecko:
Lol, just seen this. Was also working on a similar prebuildify update here #2368 

Feel free to steal bits and close as you see fit

PR #2406, opened 2022-02-01
feat: typescript everything

reconbot:
This is a huge pr, trying to get anything to convert to typescript piecemeal was very very frustrating and most of the conversions are straight copies.

However
- Big changes to `@serialport/stream`
- Big resultant downstream changes to `serialport`
- Hardware tests now only require a loopback port and they all pass (from bindings-cpp up to stream, still working on serialport)

Full change log
- Typscript types are now available for all packages
- Serialport no longer supports Node 10
- `bindings-abstract` is now `bindings-interface` and it's in it's own repo (`bindings-cpp` relies on it and having a cyclic dependency on the monorepo was hard)
- `baudRate` on `serialport` and `@serialport/stream` packages is no longer defaulted to 9600
- All packages now have named exports
- Bindings have moved from `@serialport/bindings` to `@serialport/bindings-cpp` they are shipped with - `prebuildify` and no longer require a download.
- `@serialport/bindings-cpp` now leverages N-API and shouldn't need to be upgraded for every node release.
- baudRate is now required
- bindings and mock bindings now have a new interface with bindings-cpp
- stream no longer has a list method
- serialport package no longer has a static bindings property on it's class (now a subsclass of SerialPortStream)


packages/list/lib/index.ts
```diff
@@ -1,20 +1,21 @@
...
 
-function jsonl(ports) {
+function jsonl(ports: PortInfo[]) {
   ports.forEach(port => {
     console.log(JSON.stringify(port))
   })
 }
 
-const formatters = {
+const formatters: { [key: string]: undefined | ((ports: PortInfo[]) => void) } = {
```
HipsterBrown, line 18: I find the Record type helpful to break up the "delimiter soup" of object types, ideally making it clearer that `formatters` is an object with any number of string keys which have a value of undefined or a `ports` callback. 

```suggestion
const formatters: Record<string, undefined | ((ports: PortInfo[]) => void)> = {
```

packages/list/lib/index.ts
```diff
@@ -32,7 +33,7 @@ const args = program.opts()
 autoDetect()
   .list()
   .then(ports => {
-    const formatter = formatters[args.format]
+    const formatter = formatters[args.format as string]
```
HipsterBrown, line 36: `opts` can take a generic type argument for the [OptionValues](https://github.com/tj/commander.js/blob/master/typings/index.d.ts#L676) returned by the program:

```ts
const args = program.opts<{ format: string }>()
```

packages/parser-byte-length/lib/index.test.ts
```diff
@@ -18,10 +19,10 @@ describe('ByteLengthParser', () => {
 
   it('throws when not provided with a length', () => {
     assert.throws(() => {
-      new ByteLengthParser()
+      new (ByteLengthParser as any)()
```
HipsterBrown, line 22: Does `unknown` work here?

packages/parser-byte-length/lib/index.ts
```diff
@@ -30,7 +36,7 @@ class ByteLengthParser extends Transform {
     this.buffer = Buffer.alloc(this.length)
   }
 
-  _transform(chunk, encoding, cb) {
+  _transform(chunk: Buffer, _encoding: any, cb: TransformCallback) {
```
HipsterBrown, line 39:
```suggestion
  _transform(chunk: Buffer, _encoding: BufferEncoding, cb: TransformCallback) {
```

HipsterBrown (commented):
I know this is still in draft. Just a quick look. Overall changes don't look too bad, mostly annoying in the tests to assert the runtime checks 😄

packages/parser-byte-length/lib/index.test.ts
reconbot: `new ByteLengthParser(undefined as any)` works

packages/list/lib/index.ts
reconbot: `program.opts()` takes no generics nor options in command v7 - looks like v9 is out

HipsterBrown: Dang, I guess I should have checked the dep version 😅

packages/serialport/lib/index.ts
```diff
@@ -0,0 +1,9 @@
+export * from '@serialport/parser-byte-length'
+export * from '@serialport/parser-cctalk'
+export * from '@serialport/parser-delimiter'
+export * from '@serialport/parser-inter-byte-timeout'
+export * from '@serialport/parser-readline'
+export * from '@serialport/parser-ready'
+export * from '@serialport/parser-regex'
```
reconbot, line 7: we should get the rest of the parsers in here

packages/stream/tsconfig-build.json
```diff
@@ -0,0 +1,15 @@
...
+    "outDir": "dist",
+    "rootDir": "lib",
+  },
+  "exclude": [
+    "node_modules",
+    "**/*.test.ts",
+    "dist",
+  ],
+  "references": [
+    { "path": "../parser-delimiter/tsconfig-build.json"}
```
reconbot, line 13: this can go, not used

PR #2428, opened 2022-02-10
feat: platform specific open options for streams

reconbot:
Make it a bit easier to have platform specific open options. I'm in love with the solution but it doesn't seem possible to filter the generic via a typecheck and have it apply to the constructor options on a serialport stream. So casting the open options is the only recourse I've figured out.

```ts
new SerialPort({
  path: '/bad/port',
  baudRate: 9600,
  vmin: 10,
} as LinuxOpenOptions)
```

### DIFF
--- a/packages/serialport/lib/serialport.ts
+++ b/packages/serialport/lib/serialport.ts
@@ -1,11 +1,11 @@
-import { ErrorCallback, OpenOptions, SerialPortStream } from '@serialport/stream'
-import { autoDetect, AutoDetectTypes } from '@serialport/bindings-cpp'
+import { ErrorCallback, OpenOptions, SerialPortStream, StreamOptions } from '@serialport/stream'
+import { autoDetect, AutoDetectTypes, OpenOptionsFromBinding } from '@serialport/bindings-cpp'
 
 const DetectedBinding = autoDetect()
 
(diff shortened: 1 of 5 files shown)


packages/serialport/lib/serialport.ts
```diff
@@ -1,11 +1,11 @@
-import { ErrorCallback, OpenOptions, SerialPortStream } from '@serialport/stream'
-import { autoDetect, AutoDetectTypes } from '@serialport/bindings-cpp'
+import { ErrorCallback, OpenOptions, SerialPortStream, StreamOptions } from '@serialport/stream'
+import { autoDetect, AutoDetectTypes, OpenOptionsFromBinding } from '@serialport/bindings-cpp'
 
 const DetectedBinding = autoDetect()
```
HipsterBrown, line 4: Since `autoDetect` returns a non-discriminating union (`AutoDetectTypes`), I think that is why the type assertion is needed from `DetectedBinding` later. Is it possible to [brand](https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d) each binding interface to help with this detection narrowing?

reconbot: Branded types are rad but not helping in this case. I'm positive we'll revisit this.

PR #2591, opened 2023-01-31
Add StartEndParser

glenn-kroeze:
I ran into a situation where I needed to write an interface for a protocol that can basically be described as `[STX, ...data, ETX]`. I wrote it a while ago and thought it might be useful for other people to have access to.


zjee:
good!

reconbot (approved):
This is great!

Issue #2566, opened 2022-11-26
Error with @serialport/stream/dist/index.d.ts


AndrewLuGit:
### SerialPort Version

10.5.0

### Node Version

v16.16.0

### Platform

Microsoft Windows NT 10.0.22621.0 x64

### What steps will reproduce the bug?

I am working on a VSCode extension. The error occurs when trying to run the extension.

### What happens?

I get this error:
```
node_modules/@serialport/stream/dist/index.d.ts:100:5 - error TS2416: Property '_write' in type 'SerialPortStream<T>' is not assignable to the same property in base type 'Duplex'.
  Type '(data: Buffer, encoding: BufferEncoding | undefined, callback: (error: Error | null) => void) => void' is not assignable to type '(chunk: any, encoding: string, callback: (error?: Error | null | undefined) => void) => void'.
    Types of parameters 'encoding' and 'encoding' are incompatible.
      Type 'string' is not assignable to type 'BufferEncoding | undefined'.

100     _write(data: Buffer, encoding: BufferEncoding | undefined, callback: (error: Error | null) => void): void;
```

### What should have happened?

It shouldn't have gotten this error? If I manually change the line in the file to `_write(data: Buffer, encoding: BufferEncoding, callback: (error: Error | null) => void): void;`, it runs fine.

PR #2643, opened 2023-06-27
Fix encoding for stream _write

GazHank:
Fix per https://github.com/serialport/node-serialport/issues/2566

This matches the definition of _write definition for Duplex streams


reconbot (approved):

GazHank:
Hmm, do we need to look at changing the _writev function too?

https://github.com/serialport/node-serialport/blob/0585e64b7b3a353f6d5427c405ba4a0dd989563a/packages/stream/lib/index.ts#L275

pierreroth64:
Any news on this one?

PR #2646, opened 2023-06-28
Migrate to NPM workspaces

GazHank:
Allows upgrade to Lerna v7 removing it's legacy package management

**Breaking change:**
Drops support for Node 14 (since node 15+ is required for workspaces)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.config.os }}
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x, 20.x]
+        node-version: [16.x, 18.x, 20.x]
         config:
           # arch isn't used and we have no way to use it currently
           - { os: macos-latest, arch: x64 }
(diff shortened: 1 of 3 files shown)


GazHank:
I'm not sure if we are comfortable dropping node 14 just yet... although it is out of support.  It also looks like [commander ](https://github.com/serialport/node-serialport/pull/2640) also drops support for node 14...

@reconbot would you be able to take a look at this as I'm not certain how you are using Lerna at the moment so don't know if this will break anything, or change how we want to user Lerna?

GazHank:
Oh no, it looks like we might need to revert prettier, as lerna [7.1.2](https://github.com/lerna/lerna/releases/tag/v7.1.2) is the first version to support prettier v3 :-( 

I think the revert wilI be simpler at the moment than updating to lerna 7

Alternatively to update to lerna 7 I think we need to:
- drop node 14
- switch to npm workstreams - or use the [legacy package management](https://lerna.js.org/docs/legacy-package-management)
- work out what changes we need to cope with the [switch to v7](https://lerna.js.org/docs/lerna6-obsolete-options)
- and maybe some extra steps

PR #2669, opened 2023-07-26
chore(deps): add eslint-plugin-prettier v5

GazHank:


GazHank:
worth a try, but this doesnt seem to fix the current issue

PR #2670, opened 2023-07-26
chore(deps): revert prettier to 2.8.8

GazHank:


GazHank:
@reconbot can you check if applying this change allows you to publish using Lerna again?

PR #2686, opened 2023-08-23
Upgrade to Lerna v7 and drop support for Node 14

GazHank:
There were several dependencies which couldn't be updated until node 14 was dropped, additionally Lerna 7 required a few changes, so I've bundled these together to avoid breaking interdependencies.  Hopefully this can be the basis for the next release.

This includes:

- Dropping Node 14 (githhub action change, @tsconfig/node18, etc)
- Switch to NPM workspaces
- Upgrade to Lerna v7
- Upgrade to Prettier v3
- Upgrade to bindings-cpp v12
- Upgrade to commander v11
- Upgrade to enquirer v2.4.1

Please note the switch to NPM workspaces results in a single package-lock.json in the root.  The individual lock files are no longer used, but I have not deleted these yet.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.config.os }}
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x, 20.x]
+        node-version: [16.x, 18.x, 20.x]
         config:
           # arch isn't used and we have no way to use it currently
           - { os: macos-latest, arch: x64 }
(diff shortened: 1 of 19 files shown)


GazHank:
Hi @reconbot I've checked this update and it seems to work successfully for me... although I can't complete the final steps of the lerna publish due to permissions (I think)

Would you be able to take a look and publish when you get chance?

I should call out that this doesn't include changes for the Linux issues on Node v20.  I think that will require a fair bit more investigation

reconbot (approved):
This looks great I'll give it a try. I just setup a monorepo for another project with pnpm workspaces (which covers publishing) and changesets (that covers versioning and changelogs) and it works great! Something for us to consider too.

PR #2709, opened 2023-09-23
Add support for multiple delimiters in PacketLengthParser, as well as delimiters of sizes > 1 byte

bbeck13:
I've used this for a situation where the delimiter can be many different things each specific delimiter denotes a type of packet, and because there are so many it has to be multiple bytes. This doesn't break any existing functionality and consumers do not need to use this if they stick with the defaults.

I've updated the documentation for this here
https://github.com/serialport/website/pull/125


bbeck13:
https://github.com/serialport/node-serialport/discussions/2679

So there's at least one other person who wants this.

bbeck13:
> Is there any chance this will get merged?

I wish I'm not sure who to @ to move this along

I ended up building my branch into the projects I'm using this for feel free to do the same, kind of annoying bc it doesn't publish anywhere but not the end of the world

PR #2966, opened 2024-12-24
fix: linting

reconbot:
Eslint9 and stylistic mean we can ditch prettier, typescript specific rules, and a can revisit the whole eslint config.


test/assert.ts
```diff
@@ -1,4 +1,5 @@
-import { assert, use } from 'chai'
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const { assert, use } = require('chai')
```
reconbot, line 2: I think chai went to node modules and while this probably would have just worked typescript wouldn't stop complaining

PR #2970, opened 2024-12-24
add abilitiy for multiple delimitiers and multibyte delimiters to packet length parser

reconbot:
Because of the time it took to merge #2709 I've had to cherry pick the commit and fix conflicts.

closes #2709


bbeck13:
thanks @reconbot glad to see this made it in.

PR #3066, opened 2025-07-18
chore: unbreak eslint configs after eslint-plugin-mocha upgrade

rotu:


rotu:
red CI for unrelated reasons, fixed in #3065

